PR #19, opened 2026-03-21
Insert and security checks

ggela22:


crdt-core/src/document.rs
```diff
@@ -20,8 +20,77 @@ impl Document {
...
+
+        let new_block = Block::new(new_id, left_neighbor, right_neighbor, content.to_string());
+
+        let block_len = new_block.len;
+        self.store.insert(new_block);
+
+        if let Some(left_id) = left_neighbor {
+            self.store
+                .get_mut(&left_id)
+                .unwrap()
```
CarlTeapot: these unwrap() calls seem unsafe, if they panic, we could have a case where block is appended to the document and in the structstore, but state vector  is not updated

You could do something like this:

    self.store.get_mut(&left_id)
        .ok_or(DocumentError::BlockNotFound(left_id))?;

CarlTeapot (changes requested):
მადლობა გელა

ChocoChichia (changes requested):

CarlTeapot (changes requested):

PR #30, opened 2026-04-08
PeerCode: Implement document initialization and insert/delete ipc handlers and calls 

### LINKED ISSUE #22: Implement Document initialization and management in Tauri

CarlTeapot:
* Implemented Document initialization in Tauri using Tauri Mutex States 
*  Implemented IPC handlers for insert and delete in tauri. 
* Implement IPC calls in frontend
* Implement document state visualization in the terminal for testing purposes
* updated node dependencies to fix security vulnurabilities

### DIFF
--- a/crdt-core/src/document.rs
+++ b/crdt-core/src/document.rs
@@ -41,6 +41,7 @@ impl Document {
         text
     }
 
+    #[cfg(debug_assertions)]
     pub fn debug_linked_list(&self) -> String {
         let mut parts = Vec::new();
         let mut curr = self.head;
--- a/crdt-core/src/document.rs
+++ b/crdt-core/src/document.rs
@@ -41,6 +41,38 @@ impl Document {
         text
     }
 
+    pub fn debug_linked_list(&self) -> String {
+        let mut parts = Vec::new();
+        let mut curr = self.head;
+
+        while let Some(id) = curr {
+            if let Some(block) = self.store.get(&id) {
+                let content = if block.content().is_empty() {
+                    "<empty>".to_string()
+                } else {
+                    block.content().replace('\n', "\\n")
+                };
+
+                if block.is_deleted {
+                    parts.push(format!("[DEL:{content}]"));
+                } else {
+                    parts.push(content);
+                }
+
+                curr = block.right();
+            } else {
+                parts.push("<broken-link>".to_string());
+                break;
+            }
+        }
+
+        if parts.is_empty() {
+            "<empty>".to_string()
+        } else {
+            parts.join(" --- ")
+        }
+    }
+
     fn find_insert_position(&self, block: &Block) -> Result<Option<BlockId>, DocumentError> {
         let mut left = block.origin_left;
         let right = block.origin_right;
--- a/tauri-app/src-tauri/src/crdt_handler.rs
+++ b/tauri-app/src-tauri/src/crdt_handler.rs
@@ -1,0 +1,27 @@
+use tauri::State;
+
+use crate::appstate::AppState;
+
+#[tauri::command]
+pub fn insert(state: State<AppState>, position: u64, content: String) -> Result<(), String> {
+    let mut document = state
+        .document
+        .lock()
+        .map_err(|_| "failed to lock document state".to_string())?;
+
+    document
+        .local_insert(position, &content)
+        .map_err(|err| format!("insert failed: {err:?}"))
+}
+
+#[tauri::command]
+pub fn delete(state: State<AppState>, position: u64, length: u64) -> Result<(), String> {
+    let mut document = state
+        .document
+        .lock()
+        .map_err(|_| "failed to lock document state".to_string())?;
+
+    document
+        .delete(position, length)
+        .map_err(|err| format!("delete failed: {err:?}"))
+}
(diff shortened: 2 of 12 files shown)


tauri-app/src-tauri/src/lib.rs
```diff
@@ -7,8 +33,17 @@ fn greet(name: &str) -> String {
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
     tauri::Builder::default()
+        .setup(|app| {
+            app.manage(AppState::new(ClientId::new(random::<u64>())));
```
ggela22, line 33: worth thinking about what happens if someone disconnects and reconnects. would it be better to persist the id to disk?

tauri-app/src/App.tsx
```diff
@@ -21,47 +22,74 @@ function App() {
...
+            }
+
+            const wireMessage = JSON.stringify({
+              type: opType,
+              offset,
+              ...(deleteLen > 0 && { length: deleteLen }),
+              ...(insertText.length > 0 && { text: insertText }),
+            });
+
+            try {
```
ggela22, line 75: this might cause problems if a remote operation integrates between the delete and insert. would probably be better to add a single "replace" tauri command that keeps the same mutex lock between both operations

tauri-app/src-tauri/src/lib.rs
```diff
@@ -1,4 +1,30 @@
...
+use std::thread;
+use std::time::Duration;
+
+use crate::appstate::AppState;
+use crdt_core::types::ClientId;
+use rand::random;
+use tauri::Manager;
+
+fn spawn_linked_list_logger(app_handle: tauri::AppHandle) {
+    thread::spawn::<_, ()>(move || loop {
```
ChocoChichia, line 15: this thread is never stopped properly. It loops endlessly without any way to shut down when the app closes.

ggela22 (changes requested):

tauri-app/src-tauri/src/lib.rs
```diff
@@ -1,4 +1,30 @@
...
+
+use std::thread;
+use std::time::Duration;
+
+use crate::appstate::AppState;
+use crdt_core::types::ClientId;
+use rand::random;
+use tauri::Manager;
+
+fn spawn_linked_list_logger(app_handle: tauri::AppHandle) {
```
ChocoChichia, line 14: I think we would still need this fn later so lets fix some problems   
it would be better practice to use rusts #[cfg(debug_assertions)] . polling log which prints entire linked list for prod does not sound good.

tauri-app/src-tauri/src/lib.rs
```diff
@@ -1,4 +1,30 @@
...
+    thread::spawn::<_, ()>(move || loop {
+        let state = app_handle.state::<AppState>();
+        match state.document.lock() {
```
ChocoChichia: this may cause problems for big documents later as it could be locked for a long time

tauri-app/src-tauri/src/crdt_handler.rs
ChocoChichia, line 1: you could also add get_text , so that frontend will be able to read the initial load and so on.

crdt-core/src/document.rs
```diff
@@ -41,6 +41,38 @@ impl Document {
         text
     }
 
+    pub fn debug_linked_list(&self) -> String {
```
ChocoChichia, line 45: this fn is pub on a pub struct, which is bad as  this is a crdt-core library and this fn should not be leaked to public. You decide, You can choose between #[cfg(debug_assertions)] or  pub(crate)

tauri-app/src-tauri/src/lib.rs
```diff
@@ -7,8 +33,17 @@ fn greet(name: &str) -> String {
...
     tauri::Builder::default()
+        .setup(|app| {
+            app.manage(AppState::new(ClientId::new(random::<u64>())));
+            spawn_linked_list_logger(app.handle().clone());
+            Ok(())
+        })
         .plugin(tauri_plugin_opener::init())
-        .invoke_handler(tauri::generate_handler![greet])
+        .invoke_handler(tauri::generate_handler![
+            greet,
```
ChocoChichia: kill this guy

tauri-app/src-tauri/src/lib.rs
ChocoChichia, line 33: jjust a heads up: generating a random ClientId on startup means the CRDT will treat us as a completely new user on every app restart. This will bloat the state vector over time.

if not now, I guess we can add disk persistence later ; p

ChocoChichia (changes requested):

tauri-app/src-tauri/src/crdt_handler.rs
CarlTeapot: Thats a separate task

tauri-app/src-tauri/src/lib.rs
CarlTeapot: I stubbed this as a temporary solution. correctly handling client id generation, persistence and type should be an another task

tauri-app/src-tauri/src/lib.rs
CarlTeapot: if we use #[cfg(debug_assertions)], that means this won't be used in prod. This is just for testing purposes

ChocoChichia (approved):

PR #41, opened 2026-04-18
PeerCode: Implement websocket communication system in tauri and gateway

### LINKED ISSUE #25: Research effective WebSocket libraries for Go and Rust and implement basic websocket connection between the two

CarlTeapot:
* Implemented websocket data types and states in tauri
*  Implemented immediate connection establishment with gatewway after processes are spawned
*  Implemented connection sender and receiver functions
* added connection timeout value as a configuration
* added a configuration file and its handler in lib.rs


tauri-app/src-tauri/src/state/ws_state.rs
```diff
@@ -0,0 +1,101 @@
...
+        let mut guard = self.connection.lock().await;
+        match &mut *guard {
+            WsConnection::Connected { sink, .. } => sink
+                .send(msg)
+                .await
+                .map_err(|e| WsError::SendFailed(e.to_string())),
+            _ => Err(WsError::NotConnected),
+        }
+    }
+}
```
ChocoChichia: TCP has backpressure: if the receive buffer on the other side is full,      
 sink.send() will await for potentially hundreds of milliseconds, or longer
For that entire duration, the connection mutex is locked                    
Any other code path that needs the connection (future: disconnect,
  reconnect, health check) is completely blocked waiting for the mutex           
                                                                                 
how can you fix with present arch :                                                           
  Change sink: Sink to sink: Arc<Mutex<Sink>> inside WsConnection::Connected. In
  send():                                                                        
  1. Lock self.connection, extract a clone of the Arc<Mutex<Sink>>, then drop the
   connection lock immediately                                                   
  2. Lock just the sink and do the await outside the connection lock
                                                                                 
my recommendation : 
now or later use new mpsc instead of mutex

tauri-app/src-tauri/src/state/ws_state.rs
```diff
@@ -0,0 +1,101 @@
...
+
+impl WsState {
+    pub fn new(connect_timeout: Duration) -> Self {
+        Self {
+            connection: Mutex::new(WsConnection::Disconnected),
+            connect_timeout,
+        }
+    }
+
+    pub async fn connect(&self, url: &str, session_id: String) -> Result<(), WsError> {
```
ChocoChichia, line 28: when  Gateway process restarts, or network drops, or server sends a Close frame, receive_loop hits the break and the function returns . The JoinHandle stored in reader is now finished,  but nobody updates        
  WsConnection , add a temp dead code for now.                                                                  
  receive_loop needs a reference back to the shared connection state so it can   
  -  reset it to Disconnected when it exits.
  - In connect(), clone the Arc and pass it into receive_loop as a parameter
  - At the very end of receive_loop (after the while loop exits for any reason), 
  lock the arc and set state to Disconnected

gateway/internal/hub/hub.go
```diff
@@ -1,14 +1,56 @@
...
+
+		if err := conn.Write(ctx, msgType, msg); err != nil {
+			fmt.Printf("[gateway] write error room=%s: %v\n", roomID, err)
+			return
+		}
+	}
+}
+func gracefulClose(ctx context.Context, conn *websocket.Conn, reason string) {
+	_ = conn.Close(websocket.StatusNormalClosure, reason)
 }
```
ChocoChichia, line 59: the gracefulClose function exists but is never
wired up. So clients never receive the proper Close handshake, from the      
client's perspective the connection just drops, which looks identical to a
network error.

  Fix guide: Before each return in the read/write error paths, call              
  gracefulClose(ctx, conn, "") first. The defer conn.CloseNow() stays as a safety
   fallback for panics and unhandled paths. Pattern:                             
  if err != nil { 
      gracefulClose(ctx, conn, "")
      fmt.Printf("... disconnected ...")                
      return
   Close()                                                                       
  }

tauri-app/src-tauri/src/state/ws_state.rs
ChocoChichia, line 1: Architectural note (ws_state.rs): The current design uses Mutex<WsConnection>
for shared access. The idiomatic async Rust alternative is a dedicated owner
task: give that task exclusive ownership of the sink/stream, and communicate
with it via an mpsc::Sender<Message>. No mutex needed, the channel is the  
synchronisation. Worth considering before this grows, because the            
mutex-across-await problem disappears entirely with this pattern.

ChocoChichia (changes requested):

ChocoChichia (approved):

PR #45, opened 2026-04-20
44 refactor crdt library

### LINKED ISSUE #44: refactor crdt library
1. Fix `origin_left` to reference the exact character clock (not block start) after mid-block splits.
2. Add remote split phase so origins never point inside unsplit multi-char blocks.
3. Correct YATA scan to skip descendants instead of breaking on any origin mismatch.
4. Apply client-ID tiebreaker solely on `origin_left`, removing incorrect `origin_right` dependency.
5. Maintain sorted block order in `StructStore` (no unconditional `push`) to preserve binary search correctness.
6. Properly reject out-of-bounds positions even in empty documents.
7. Preserve original `origin_right` when splitting blocks instead of using current neighbors.
8. Buffer and reapply deletes for missing blocks instead of silently dropping them.

ChocoChichia:
## Corrected bugs below for crdt-core library

1. Fix `origin_left` to reference the exact character clock (not block start) after mid-block splits.
2. Add remote split phase so origins never point inside unsplit multi-char blocks.
3. Correct YATA scan to skip descendants instead of breaking on any origin mismatch.
4. Apply client-ID tiebreaker solely on `origin_left`, removing incorrect `origin_right` dependency.
5. Maintain sorted block order in `StructStore` (no unconditional `push`) to preserve binary search correctness.
6. Properly reject out-of-bounds positions even in empty documents.
7. Preserve original `origin_right` when splitting blocks instead of using current neighbors.
8. Buffer and reapply deletes for missing blocks instead of silently dropping them.
9. Consolidate `DocumentError` into `error.rs`, removing the stale duplicate definition in `types.rs`. Add `Display` + `std::error::Error` impls and a new `PendingQueueFull` variant for future queue-bound enforcement. Re-export at crate root so `crdt_core::DocumentError` works.
10. Classify every incoming remote block as `Ready` / `Pending` / `Duplicate`; drop duplicates silently in both `remote_insert` and `drain_pending`. Previously duplicates slipped past `StateVector::can_integrate` (which only matched `seen == clock`) and stayed in `pending_blocks` forever - a silent memory leak on any retransmit.
11. Make `delete()` return `OutOfBounds` instead of silently no-op'ing. A missing start block returns `Err(OutOfBounds(position))`; an over-delete (e.g. `delete(0, 999)` on a 5-char doc) returns `Err(OutOfBounds(first_unreachable_position))` while still preserving the valid prefix that was already marked deleted.
12. Add a debug-only cycle guard to `get_text` and `debug_linked_list`. A corrupted or hostile peer's self-referencing block would previously infinite-loop the render thread; now `debug_assert!` fires and `total_blocks()` on `StructStore` provides the bound. Zero cost in release builds via `#[cfg(debug_assertions)]`.

## Refactored and split `document.rs` into focused submodules

13. Break the ~600-line `document.rs` into four files under `src/document/`, each scoped to one responsibility:
    - `traversal.rs` -  `get_text`, `debug_linked_list`, `get_block_and_offset_by_position`.
    - `integrate.rs` - YATA scan, `link_block`, `integrate`, `split_block`, `pre_split_for_block`.
    - `ops.rs` - `resolve_origins`, `local_insert`, `delete`, `collect_garbage`.
    - `pending.rs` -`Readiness` enum, `classify_block`, `remote_insert`, `drain_pending`, `try_apply_delete_set`, `apply_delete_set`.
14. added  more test coverage

## Known follow-ups (not in this PR)

- Networking surface for Tauri: `local_insert` / `delete` should return the `Block` / `DeleteSet` delta; need `snapshot()`, `state_vector()`, `apply_update(blocks, ds)`, `encode_update(since)`.
- Enforce `PendingQueueFull` cap (variant exists, not yet thrown).
- Wire format (`serde` feature) for `Block` / `DeleteSet` / `StateVector`.
- Basic input validation on `remote_insert` (len matches content, origins point at known clients, etc.).

#44


crdt-core/src/error.rs
```diff
@@ -1,19 +1,25 @@
...
 
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum DocumentError {
     BlockNotFound(BlockId),
+    OutOfBounds(u64),
+    PendingQueueFull,
 }
 
 impl std::fmt::Display for DocumentError {
```
CarlTeapot: What do you think about getting rid of the full dependency paths when its not absolutely necessary to specify which dependency it is (in this case, Display is a well known attribute and its safe to assume that it means std::fmt::Display)

crdt-core/src/store/struct_store.rs
```diff
@@ -86,13 +81,4 @@ impl StructStore {
...
-
-    // Empty for now , will implement later when we have the basic structure in place
-    //pub fn get_missing_blocks(&self, remote_sv: &StateVector) -> Vec<&Block> {    }
-
-    //pub fn split_block_at(&mut self, client: ClientId, split_clock: u64) -> Option<BlockId> {}
-
-    //pub fn try_squash_tail(&mut self, client: ClientId) {}
-
-    //pub fn undelete(&mut self, id: &BlockId) -> bool {}
 }
```
CarlTeapot, line 84: kargia

crdt-core/src/store/struct_store.rs
```diff
@@ -18,19 +18,14 @@ impl StructStore {
         self.get(id).is_some()
     }
 
-    pub fn insert(&mut self, block: Block) {
-        let list = self.blocks.entry(block.id.client).or_default();
-        list.push(block);
+    pub fn total_blocks(&self) -> usize {
+        self.blocks.values().map(|v| v.len()).sum()
     }
```
CarlTeapot, line 23: this could also be O(1) by making it a global variable. but it will need careful management. If you think that performance gain will be impactful, then we should do this. your choice

crdt-core/src/document/pending.rs
```diff
@@ -0,0 +1,164 @@
+use super::Document;
+use crate::error::DocumentError;
+use crate::store::DeleteSet;
+use crate::structs::Block;
+use crate::types::{BlockId, Clock};
+
+/// Classification of an incoming remote block relative to local state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Readiness {
```
CarlTeapot: what do you think about naming this "Insert_Readiness" or "Block_readiness" or something else? We can do so that enum name combined with state names can be pretty self descriptive and the comments won't be needed

crdt-core/src/document/pending.rs
```diff
@@ -0,0 +1,164 @@
...
+            }
+            Readiness::Ready => {}
+        }
+
+        let client = block.id.client;
+        let end_clock = block.id.clock.value + block.len;
+
+        self.pre_split_for_block(&block)?;
+        self.integrate(block)?;
+        self.state_vector.update(client, end_clock);
```
CarlTeapot, line 59: you call this here and in drain_pending. It is relatively short and I don't think this needs a separate function, but if these three lines are tightly integrated and you should't call these functions without the other, then maybe you should extract it in a separate function and enforce that the document state is updated with that function

crdt-core/src/document/pending.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn classify_block(&self, block: &Block) -> Readiness {
+        let seen = self.state_vector.get(&block.id.client);
+        let clock = block.id.clock.value;
+
+        if seen > clock {
+            return Readiness::Duplicate;
+        }
+        if seen < clock {
+            return Readiness::Pending;
+        }
```
CarlTeapot, line 74: you could also do: 

    match seen.cmp(&clock) {
        Ordering::Greater => Readiness::Duplicate,
        Ordering::Less    => Readiness::Pending,
        Ordering::Equal   => {
         ......
         }

your code is more intuitive and easier to read, but this seems more "Rusty". your choice

crdt-core/src/document/pending.rs
```diff
@@ -0,0 +1,164 @@
...
+            let end_clock = range.end();
+
+            while current_clock < end_clock {
+                let id = BlockId::new(*client, Clock::new(current_clock));
+
+                let (block_start, block_len, block_id) = match self.store.get(&id) {
+                    Some(b) => (b.id.clock.value, b.len, b.id),
+                    None => {
+                        unapplied.add(
+                            BlockId::new(*client, Clock::new(current_clock)),
```
CarlTeapot: why not just use "id" variable  defined in the while block?

crdt-core/src/document/pending.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        let client = block.id.client;
+        let end_clock = block.id.clock.value + block.len;
+
+        self.pre_split_for_block(&block)?;
+        self.integrate(block)?;
+        self.state_vector.update(client, end_clock);
+
+        self.drain_pending()?;
+        Ok(())
```
CarlTeapot: Every case returns Ok(()). This means that Tauri client won't be able to figure out what happened and won't know if the frontend needs to be updated or not. We should return a proper response here either with a specific error type or an enum. We will also need to return the position index and the block content as a string, so that tauri can emit the remote changes to the monaco frontend

crdt-core/src/document/pending.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    pub fn apply_delete_set(&mut self, remote: &DeleteSet) -> Result<(), DocumentError> {
+        let unapplied = self.try_apply_delete_set(remote)?;
+        if !unapplied.is_empty() {
+            self.pending_delete_sets.push(unapplied);
+        }
+        self.seen_delete_set.merge(remote);
+        Ok(())
+    }
```
CarlTeapot: We could get rid of the emptyness check if the `try_apply_delete_set `returned an error when the unapplied is empty. I am not sure if that is a good way to write rust code, but this could be made shorter that way. your choice bro

crdt-core/src/document/pending.rs
```diff
@@ -0,0 +1,164 @@
...
+        Readiness::Ready
+    }
+
+    /// Repeatedly drain pending blocks and pending delete sets until a pass
+    /// makes no further progress.
+    fn drain_pending(&mut self) -> Result<(), DocumentError> {
+        loop {
+            let mut progress = false;
+
+            let candidates: Vec<Block> = self.pending_blocks.drain(..).collect();
```
CarlTeapot: Use` mem::take` here.
`drain(..).collect() `allocates new memory and moves data into it.  
`mem::take` just swaps out the pointer.

crdt-core/src/document/ops.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// Reclaim storage for every block whose tombstone is covered by
+    /// `confirmed`. Content bytes are cleared
+    pub fn collect_garbage(&mut self, confirmed: &DeleteSet) {
+        for (client, range) in confirmed.iter() {
+            let mut current_clock = range.start;
+            let end_clock = range.end();
+
+            while current_clock < end_clock {
+                let id = BlockId::new(*client, crate::types::Clock::new(current_clock));
```
CarlTeapot: the crate is already imported bro

crdt-core/src/document/ops.rs
```diff
@@ -0,0 +1,144 @@
...
+                    .ok_or(DocumentError::BlockNotFound(block_id))?;
+                let origin_left_id =
+                    BlockId::new(block_id.client, block_id.clock.advance(offset - 1));
+                Ok((Some(origin_left_id), left_ref.right()))
+            }
+        } else if offset > 0 {
+            Err(DocumentError::OutOfBounds(position))
+        } else {
+            Ok((tail_id, None))
+        }
```
CarlTeapot, line 27: you could invert this conditional to reduce nesting: 
```

let (block, offset, tail_id) = self.get_block_and_offset_by_position(position);

    let Some(block_id) = block else {
        if offset > 0 {
            return Err(DocumentError::OutOfBounds(position));
        }
        return Ok((tail_id, None));
    };

    if offset == 0 {
        let block_ref = self
            .store
            .get(&block_id)
            .ok_or(DocumentError::BlockNotFound(block_id))?;
        return Ok((block_ref.left(), Some(block_id)));
    }

    self.split_block(block_id, offset)?;
    let left_ref = self
        .store
        .get(&block_id)
        .ok_or(DocumentError::BlockNotFound(block_id))?;
    let origin_left_id = BlockId::new(block_id.client, block_id.clock.advance(offset - 1));
    Ok((Some(origin_left_id), left_ref.right()))
```

CarlTeapot (changes requested):

PR #52, opened 2026-04-22
49 define crdt wire protocol

### LINKED ISSUE #49: Define CRDT wire protocol
Pick serde format (`postcard` strongly suggested — compact, no_std, 0 unsafe). Add an envelope: `enum OpMessage { Insert(Block), Delete(DeleteSet), Snapshot(DocumentSnapshot) }`. Derive `Serialize, Deserialize` on `Block`, `DeleteSet`, `BlockId`, `ClientId`, `Clock`.

ChocoChichia:
## Define CRDT wire protocol (#49)

Add a binary framing layer so CRDT ops can travel over WebSocket between the editor and the relay.

**Rust (`crdt-core`)**
- New `wire` module: `WireBlock`, `OpMessage { Insert, Delete }`, `encode_op` / `decode_op` with a 1-byte prefix (`OP_PREFIX = 0x00`, `SNAPSHOT_PREFIX = 0x01` reserved).
- `bitcode = "0.6.9"` added; `Encode`/`Decode` derived on primitive CRDT types. `Block` excluded - runtime fields must not hit the wire.
- `local_insert` now returns `Option<WireBlock>`; `delete` returns the `DeleteSet` diff for that call. Both give the tauri layer what it needs to send without re-walking the store (T10 hook).

**Go (`gateway/internal/wire`)**
- `DecodeOpFrame` / `EncodeOpFrame`: read the 1-byte prefix, return the opaque bitcode payload. Zero allocations on the happy path. No bitcode dependency - the gateway stays a dumb relay.
- Hub validates binary frames instead of blind-echoing; malformed frames are dropped without disconnecting the room.

**Both sides**
- Bidirectional drift guard: a failing test on each side if `OP_PREFIX` or `SNAPSHOT_PREFIX` diverge.


gateway/internal/wire/wire.go
```diff
@@ -0,0 +1,38 @@
...
+
+func DecodeOpFrame(frame []byte) ([]byte, error) {
+	if len(frame) == 0 {
+		return nil, ErrEmptyFrame
+	}
+	switch frame[0] {
+	case PrefixOp:
+		return frame[1:], nil
+	case PrefixSnapshot:
+		return nil, ErrSnapshotNotSupported
```
CarlTeapot, line 29: lets add a TODO comment here

crdt-core/src/wire.rs
```diff
@@ -0,0 +1,94 @@
...
+            }
+            WireError::NotAnOp => {
+                write!(f, "wire frame carries a snapshot, not an op")
+            }
+            WireError::Decode(e) => write!(f, "bitcode decode failed: {e}"),
+        }
+    }
+}
+
+impl std::error::Error for WireError {
```
CarlTeapot: broooooooooooo

crdt-core/src/wire.rs
```diff
@@ -0,0 +1,94 @@
...
+    frame.extend_from_slice(&payload);
+    frame
+}
+
+pub fn decode_op(frame: &[u8]) -> Result<OpMessage, WireError> {
+    match frame.first() {
+        None => Err(WireError::EmptyFrame),
+        Some(&OP_PREFIX) => bitcode::decode(&frame[1..]).map_err(WireError::Decode),
+        Some(&SNAPSHOT_PREFIX) => Err(WireError::NotAnOp),
+        Some(&b) => Err(WireError::UnknownPrefix(b)),
```
CarlTeapot: Todo here too bro

crdt-core/src/wire.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
```
CarlTeapot: you could split the frame and then match:

let (&prefix, payload) = frame.split_first().ok_or(WireError::EmptyFrame)?;
match prefix {
OP_PREFIX       => bitcode::decode(payload).map_err(WireError::Decode),
SNAPSHOT_PREFIX => Err(WireError::NotAnOp),
b               => Err(WireError::UnknownPrefix(b)),
}

crdt-core/src/wire.rs
```diff
@@ -0,0 +1,94 @@
...
+
+pub const OP_PREFIX: u8 = 0x00;
+pub const SNAPSHOT_PREFIX: u8 = 0x01;
+
+#[derive(Debug, Clone, PartialEq, Eq, bitcode::Encode, bitcode::Decode)]
+pub struct WireBlock {
+    pub id: BlockId,
+    pub origin_left: Option<BlockId>,
+    pub origin_right: Option<BlockId>,
+    pub content: String,
```
CarlTeapot, line 16: because this is immutable once initialized, you could use Box<<str>> which is more memory efficient i think

CarlTeapot (dismissed):

crdt-core/src/wire.rs
ChocoChichia: vibed too hard

PR #55, opened 2026-04-24
54 gateway room registry and client room mapping

### LINKED ISSUE #37: Move Room generation logic to Gateway and introduce client->Room mapping
Currently room id is generated by tauri client. This should be handled by the gateway

### LINKED ISSUE #54: Gateway: room registry and client->room mapping

ChocoChichia:
Implements T03. Hub owns rooms under a sync.Mutex; each Room runs in its
own goroutine and exits via a done channel so HandleWS can send to      
r.Ops() without racing the close. Client has buffered send + Read/Write 
pumps ready for T04 fan-out. POST /rooms mints an 8-hex ID via          
crypto/rand so the host no longer generates room IDs locally.           
                                                                        
HandleWS now requires both ?room= and ?client_id= query params; frames  
are decoded via wire.DecodeOpFrame and forwarded to the room (fan-out   
itself is TODO(T04)).                                                   
                               
Host no longer generates room IDs locally; after the gateway reports its
port, launch() POSTs /rooms and uses the returned ID. The local loopback
ws URL now carries &client_id=<host> so the gateway can register the    
host the same way it registers guests.                                  
                                                                        
fetch_room_id is split into gateway_http_post_empty (IO) and            
parse_room_id_response (pure parser), wrapped in a 5s timeout so a      
stalled gateway can't hang host start-up.                               
                                                                        
Removes dead WsState::send and WsError::SendFailed (re-added by T10     
when actually needed).

### DIFF
--- a/gateway/cmd/server/main.go
+++ b/gateway/cmd/server/main.go
@@ -26,6 +26,7 @@ func main() {
 
 	mux := http.NewServeMux()
 	mux.HandleFunc("/ws", h.HandleWS)
+	mux.HandleFunc("/rooms", h.HandleCreateRoom)
 	mux.HandleFunc("/health", func(w http.ResponseWriter, _ *http.Request) {
 		w.WriteHeader(http.StatusOK)
 	})
--- a/gateway/internal/client/client.go
+++ b/gateway/internal/client/client.go
@@ -1,28 +1,72 @@
 package client
 
-// Client represents a single connected WebSocket peer.
-//
-// Responsibilities:
-//   - Own the websocket.Conn
-//   - reads frames from the wire, decodes them
-//     into protocol.Message, forwards to the Room's ops channel
-//   - drains the send channel, writes frames to the wire
-//   - Handle graceful disconnect: close send channel → writePump exits →
-//     connection closed → Room is notified via leave channel
-//
-// A Client never talks to other Clients directly.
-// All cross-client communication goes through the Room's fan-out channel.
-//
-// The send channel is buffered (e.g. 256 messages).
-// If it fills up (slow reader), the client is dropped — no backpressure
-// allowed to stall the fan-out goroutine.
-//
-// Fields:
-//   ID       string            — unique per connection (uuid or random hex)
-//   RoomID   string            — which room this client belongs to
-//   IsHost   bool              — true if this client created the room
-//   conn     *websocket.Conn
-//   send     chan []byte        — outbound frame queue
-//   room     *room.Room        — back-reference for leave notification
+import (
+	"context"
 
-type Client struct{}
+	"github.com/coder/websocket"
+)
+
+const sendBufferSize = 256
+
+type Client struct {
+	ID     string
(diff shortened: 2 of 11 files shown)


gateway/cmd/server/main.go
```diff
@@ -26,6 +26,7 @@ func main() {
 
 	mux := http.NewServeMux()
 	mux.HandleFunc("/ws", h.HandleWS)
+	mux.HandleFunc("/rooms", h.HandleCreateRoom)
```
CarlTeapot, line 29: lets rename "h" to something more understandable

gateway/internal/client/client.go
```diff
@@ -1,28 +1,72 @@
...
+		ID:     id,
+		RoomID: roomID,
+		conn:   conn,
+		send:   make(chan []byte, sendBufferSize),
+	}
+}
+
+func (c *Client) CloseSend() {
+	close(c.send)
+}
```
CarlTeapot, line 29: This can panic when closing an already closed channel, I suggest this:

func (c *Client) CloseSend() {
c.closeOnce.Do(func() { close(c.send) })
}

gateway/internal/client/client.go
```diff
@@ -1,28 +1,72 @@
...
+			return
+		}
+	}
+}
+
+// drains the send channel to the websocket until it is closed
+func (c *Client) WritePump(ctx context.Context) {
+	defer func() {
+		_ = c.conn.Close(websocket.StatusNormalClosure, "")
+	}()
```
CarlTeapot, line 57: What if its not a normal closure and for example there was an error writing to the channel or something bad?

gateway/internal/hub/hub.go
```diff
@@ -1,64 +1,150 @@
...
-			}
-			fmt.Printf("[gateway] op    room=%s  payload=%d bytes\n", roomID, len(payload))
-
-			if err := conn.Write(ctx, msgType, msg); err != nil {
-				fmt.Printf("[gateway] write error room=%s: %v\n", roomID, err)
-				return
-			}
 		}
 	}
 }
```
CarlTeapot, line 150: This could be decomposed into separate functions

gateway/internal/client/client.go
```diff
@@ -1,28 +1,72 @@
...
+	for {
+		_, data, err := c.conn.Read(ctx)
+		if err != nil {
+			return
+		}
+		select {
+		case ops <- data:
+		case <-ctx.Done():
+			return
+		}
```
CarlTeapot, line 49: As I understand, the flow is the following:
the client struct holds the connection. when the client sends a message, the read pump sends data to the operation relay handler, which relays it, but in this way, how will the handler know which client sent the data through the channel so that he doesnt relay the operation back to that client? I think we can solve this by sending a struct to the channel,  struct that has data and client fields (client will be populated by readpump, not the network message)

CarlTeapot (changes requested):

PR #56, opened 2026-04-24
Implement session join and refactor process spawning

### LINKED ISSUE #24: Implement Session creation/join  UI in Monaco

### LINKED ISSUE #42: Decouple sidecar Process spawning workflow
Right now, the process spawning workflow in tauri is tightly coupled: At first, the gateway binary is being run, which, before returning, spawns the tunnel process which then calls websocket handler to establish local connection. if one of these steps fail, the entire workflow may fail and the application can not run. 

To increase readability, make error handling and lifecycle management easier, the goal is to untroduce the coordinator which spawns separate workflows and calls the WS connection manager to establish the connection

### LINKED ISSUE #53: Implement Session join feature

CarlTeapot:
* Implemented a feature to join a remote session (works!!!!)
* Implemented a frontend popup that asks the user to either host a session or join
* decoupled the process spawning: 
    * moved gateway spawning and tunnel spawning into separate files
    * introduced a process coordinator which spawns both processes separately
    * process coordinator spawns log capturing for both processes 
    * moved host->gateway websocket connection establishment to process coordinator
    * made gateway and tunnel log capturing configurable in peercode.config.toml at compile time 
    * introduced a guest client state
    * hosting and joining session will no longer be possible if the client is not in a idle or starting state
* restructured the tauri project to separate files into directories
* made frontend port configurable to allow more than one instance of peercode running on a single machine
* refactored fetch_room http call to use a library and make it shorter
* dont remember anything else but im sure there is something more
* 😜
* 😜
* 😜
* 😜


ChocoChichia:
MR.TEAPOT YOU WHERE SUPPOSED TO REFACTOR ONE METHOD

CarlTeapot:
> MR.TEAPOT YOU WHERE SUPPOSED TO REFACTOR ONE METHOD

😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜

tauri-app/src/sessionSetup.tsx
ChocoChichia, line 1: you know what you did here I am not gonna embarrass you any more

tauri-app/src-tauri/src/processes/gateway_process.rs
```diff
@@ -0,0 +1,122 @@
...
+async fn on_gateway_ready(
+    app: &AppHandle,
+    port: u16,
+    room_id: &str,
+    log_rx: Receiver<CommandEvent>,
+) -> Option<GatewayWorkflowResult> {
+    let lan_url = get_lan_url(port, room_id)
+        .await
+        .expect("LAN URL cannot be null");
+
```
ChocoChichia: get_lan_url still panics on offline / no-LAN-route hosts

returns Err on hosts with no usable interface (no Wi-Fi/Ethernet, weird VPN, some sandboxes). before your pr we threaded Option<String>. Now any of those conditions crash the entire app.

GatewayReadyPayload. lan_url is even still typed Option<String> (session_types.rs:9), which is   
the giveaway: you wrap Some(lan_url) because the field is optional, but you removed the only
place it could legitimately be None.

tauri-app/src-tauri/src/meeting_gathering_assembly_conference_sitting_workshop_dedistyvna/session.rs
```diff
@@ -0,0 +1,135 @@
...
+    );
+
+    ws.connect(&ws_url, join_info.room_id.clone())
+        .await
+        .map_err(|e| {
+            *state.role.lock().unwrap() = AppRole::Undecided;
+            e.to_string()
+        })?;
+
+    *state.role.lock().unwrap() = AppRole::Guest {};
```
ChocoChichia: on lines 35-41 you have :                                                                           
                  
  {                                                                                               
      let mut role = state.role.lock().unwrap();
      if !role.can_initiate_session() { ... }
      *role = AppRole::Starting;                                                                  
  }
  // ... (lock released) async work, then ws.connect().await ...                                  
  *state.role.lock().unwrap() = AppRole::Guest {};   // line 60 — unconditional                

so this line here is unconditional.
explicit teardown or a future cancecl command can set role to udnecided and we try to set it to guest.
do same logic as you have on host .   re-acquire the lock and only set Guest if *role is     
still Starting; otherwise drop the connection (call ws.disconnect()).

tauri-app/src-tauri/src/session/command.rs
```diff
@@ -0,0 +1,135 @@
...
+    Ok(())
+}
+
+#[tauri::command]
+pub async fn join_session(
+    url: String,
+    state: State<'_, AppState>,
+    ws: State<'_, WsState>,
+) -> Result<(), String> {
+    println!("{:#?}", url);
```
ChocoChichia, line 31: remove

tauri-app/src-tauri/src/processes/process_coordinator.rs
```diff
@@ -0,0 +1,72 @@
...
+            emit_error(&app, msg.clone());
+            return Err(msg);
+        }
+    };
+
+    if logging.show_gateway_logs {
+        tauri::async_runtime::spawn(pipe_process_logs("gateway", gateway.log_rx));
+    }
+
+    let _ = app.emit(
```
ChocoChichia: 1. emit GATEWAY_READY (App.tsx flips status to "host" and stores lan_url).                      
2. await run_cloudflared (seconds; can fail).                                                   
3. on cloudflared error -> emit_error -> SESSION_ERROR -> start_host_session returns Err ->modal
shows error.                                                                                    
                                                                                          
Frontend transiently sees "host ready" before "error", and lan_url/room_id get into UI state for
a session that's about to be torn down. Carried over from prior review.                        

consolidate into a single SESSION_READY event emitted only after both processes  
are up, OR explicitly document GATEWAY_READY as provisional and have the frontend gate UI on
TUNNEL_READY/SESSION_ERROR only.

tauri-app/src-tauri/src/meeting_gathering_assembly_conference_sitting_workshop_dedistyvna/session.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    let (base, query) = url.split_once('?').unwrap_or((&url, ""));
+
+    let room_id = query
+        .split('&')
+        .find_map(|kv| kv.strip_prefix("room="))
+        .map(|v| v.to_string())
+        .ok_or_else(|| "URL is missing the ?room= parameter".to_string())?;
+
+    let server_url = base
```
ChocoChichia: any URL whose pre-query path doesn't end in literal /ws round-trips as    
{base}/ws?room=..., producing nonsense paths for non-default deployments. not a problem now but also 

parsing accepts ws://?room=x (empty host) and other malformed inputs. We also added reqwest    
(which pulls url) and tungstenite already depends on url  , we can use real parser

Suggested fix: url::Url::parse(&url) , check scheme is ws/wss ,  query_pairs().find(|(k,_)| k == 
"room") , derive server_url as {scheme}://{authority}{path-without-trailing-/ws}.

tauri-app/src-tauri/src/processes/process_coordinator.rs
```diff
@@ -0,0 +1,72 @@
...
+    GatewayReadyPayload, TunnelReadyPayload, GATEWAY_READY, TUNNEL_READY,
+};
+use crate::processes::gateway_process::run_gateway;
+use crate::processes::tunnel_process::run_cloudflared;
+use crate::processes::types::emit_error;
+use tauri::{AppHandle, Emitter, Manager};
+use tauri_plugin_shell::process::CommandEvent;
+use tokio::sync::mpsc::Receiver;
+
+pub async fn launch(app: AppHandle) -> Result<(u16, String), String> {
```
ChocoChichia: this returns bare (u16, String) tuple, when here are already    
GatewayWorkflowResult and TunnelWorkflowResult structs in processes/types.rs. Add               
SessionStartResult { port, room_id } (or just return the existing GatewayWorkflowResult).

tauri-app/src-tauri/src/processes/types.rs
```diff
@@ -0,0 +1,29 @@
...
+pub struct RoomResponse {
+    pub(crate) room_id: String,
+}
+pub struct TunnelWorkflowResult {
+    pub public_url: String,
+    pub room_id: String,
+    pub log_rx: Receiver<CommandEvent>,
+}
+
+pub fn emit_error(app: &AppHandle, message: String) {
```
ChocoChichia: emit_error and RoomResponse don't belong here

tauri-app/src-tauri/src/state/appstate.rs
```diff
@@ -20,7 +20,7 @@ pub enum AppRole {
         lan_url: Option<String>,
         public_url: Option<String>,
     },
-    // add guest laterr
+    Guest {},
```
ChocoChichia: will you add fields to it?

tauri-app/src-tauri/src/session/command.rs
```diff
@@ -0,0 +1,135 @@
...
+    Ok(())
+}
+
+#[tauri::command]
+pub async fn disconnect_websocket(ws: State<'_, WsState>) -> Result<(), String> {
+    ws.disconnect().await.map_err(|e| e.to_string())
+}
+
+#[tauri::command]
+pub fn get_session_info(state: State<'_, AppState>) -> SessionInfo {
```
ChocoChichia, line 92: get_session_info returns nothing for guests

tauri-app/src-tauri/src/processes/process_coordinator.rs
```diff
@@ -0,0 +1,72 @@
...
+        Err(msg) => {
+            emit_error(&app, msg.clone());
+            return Err(msg);
+        }
+    }
+
+    Ok((port, resolved_room_id))
+}
+
+async fn pipe_process_logs(name: &str, mut rx: Receiver<CommandEvent>) {
```
ChocoChichia, line 62: this ignores Terminated. When the child crashes, exit info is dropped   silently. Add a Terminated { code, signal } arm that prints a one-liner , saves hours next time gateway dies in the wild.

ChocoChichia (changes requested):

tauri-app/src-tauri/src/state/appstate.rs
CarlTeapot: No

ggela22 (approved):

PR #63, opened 2026-04-29
implement op fan-out

### LINKED ISSUE #40: Implement Operation relay logic in Gateway

### LINKED ISSUE #62: Implement Operation relay logic in Gateway

ggela22:
- add BroadcastMsg{Sender, Data} so Run() can exclude the sender
  from fan-out without comparing string IDs
- implement Room.Run() fan-out: snapshot clients under lock, release,
  then Send() to each peer; ForceClose() on full buffer
- add Client.Send() with panic recovery for the CloseSend() race window
- add Client.ForceClose() running ws close in a background goroutine
  so a slow peer can't stall the fan-out loop
- fill in protocol.Message struct

Closes #40

### DIFF
--- a/gateway/internal/hub/hub.go
+++ b/gateway/internal/hub/hub.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net/http"
 	"sync"
+	"time"
 
 	"github.com/coder/websocket"
 
@@ -92,19 +93,26 @@ func readWSParams(w http.ResponseWriter, r *http.Request) (roomID, clientID stri
 	return roomID, clientID, true
 }
 
-func dispatchFrame(rm *room.Room, raw []byte, roomID, clientID string) {
-	payload, err := wire.DecodeOpFrame(raw)
-	if err != nil {
-		log.Printf("[gateway] drop frame room=%s client=%s: %v", roomID, clientID, err)
+func dispatchFrame(rm *room.Room, sender *client.Client, raw []byte) {
+	if _, err := wire.DecodeOpFrame(raw); err != nil {
+		log.Printf("[gateway] drop frame room=%s client=%s: %v", sender.RoomID, sender.ID, err)
 		return
 	}
-	// TODO(T04): replace the silent drop with either back-pressure
-	// (block with a short timeout) or disconnecting the slow peer.
-	// Dropping ops silently diverges CRDT state on the recipients.
+
+	msg := room.BroadcastMsg{Sender: sender, Data: raw}
+
 	select {
-	case rm.Ops() <- payload:
+	case rm.Ops() <- msg:
+		return
 	default:
-		log.Printf("[gateway] ops buffer full room=%s; dropping frame", roomID)
+	}
+
+	t := time.NewTimer(100 * time.Millisecond)
+	defer t.Stop()
+	select {
+	case rm.Ops() <- msg:
+	case <-t.C:
+		log.Printf("[gateway] ops buffer full room=%s client=%s; dropping frame", sender.RoomID, sender.ID)
 	}
 }
 
@@ -140,7 +148,7 @@ func (h *Hub) HandleWS(w http.ResponseWriter, r *http.Request) {
 	for {
 		select {
 		case raw := <-ops:
-			dispatchFrame(rm, raw, roomID, clientID)
+			dispatchFrame(rm, c, raw)
 		case <-leave:
 			return
 		case <-ctx.Done():
--- a/gateway/internal/protocol/protocol.go
+++ b/gateway/internal/protocol/protocol.go
@@ -35,4 +35,10 @@ const (
 	MsgPong     = "pong"
 )
 
-type Message struct{}
+type Message struct {
+	Type     string `json:"type"`
+	Room     string `json:"room"`
+	ClientID string `json:"client_id"`
+	Seq      uint64 `json:"seq"`
+	Payload  string `json:"payload"`
+}
--- a/gateway/internal/room/room.go
+++ b/gateway/internal/room/room.go
@@ -1,12 +1,18 @@
 package room
 
 import (
+	"log"
 	"sync"
 
 	"gateway/internal/client"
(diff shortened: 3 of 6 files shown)


gateway/internal/hub/hub.go
```diff
@@ -92,19 +93,26 @@ func readWSParams(w http.ResponseWriter, r *http.Request) (roomID, clientID stri
...
+	}
+
+	t := time.NewTimer(100 * time.Millisecond)
+	defer t.Stop()
+	select {
+	case rm.Ops() <- msg:
+	case <-t.C:
+		log.Printf("[gateway] ops buffer full room=%s client=%s; dropping frame", sender.RoomID, sender.ID)
 	}
 }
```
CarlTeapot, line 117: I think the current flow isn't perfect. The client owns the websocket connection, read pump receives the raw frame, then sends the message to this function, which receives the message chooses the room and then sends it to the room channel, which broadcasts it. I think this:
1) does not satisfy single responsibility principle: hub should be for managing room lifecycles and assigning new clients to rooms, not for relaying messages
2) it forces all room messages to go through the same channel, when they could be separated completely

what do you think about changing the architecture so that client holds the per-room channel and sends the frames there, then we have rooms relaying these messages.

gateway/internal/protocol/protocol.go
```diff
@@ -35,4 +35,10 @@ const (
...
 )
 
-type Message struct{}
+type Message struct {
+	Type     string `json:"type"`
+	Room     string `json:"room"`
+	ClientID string `json:"client_id"`
+	Seq      uint64 `json:"seq"`
+	Payload  string `json:"payload"`
+}
```
CarlTeapot, line 44: what is this message for? is it a regular operation message that gateway relays? then it doesn't need the room field, client already holds that information. it doesn't need client id either. the client id is only needed for initial client connection and relaying. what is seq variable?

gateway/internal/room/room.go
```diff
@@ -71,9 +77,22 @@ func (r *Room) Run() {
 		select {
 		case <-r.done:
 			return
```
CarlTeapot, line 80: Some messages might still be in the channel when done is received, we should drain it if thats correct

gateway/internal/room/room.go
```diff
@@ -71,9 +77,22 @@ func (r *Room) Run() {
...
+					targets = append(targets, c)
+				}
+			}
+			r.mu.Unlock()
+
+			for _, c := range targets {
+				if !c.Send(msg.Data) {
+					log.Printf("[room] %s: disconnecting slow client %s", r.ID, c.ID)
+					c.ForceClose()
+				}
```
CarlTeapot: when this happens and the connection is dead, before the client reconnects, we will still be sending all ops to the client for no reason. We should keep the client in the map in case he reconnects, but we could introduce a state that says if the client is connected or disconnected and avoid relaying the messages to disconnected peers and do atomic state transitions. we can ignore reconnect for now but do that maybe in disconnect

gateway/internal/room/room.go
```diff
@@ -71,9 +77,22 @@ func (r *Room) Run() {
...
+			}
 		}
 	}
 }
```
CarlTeapot, line 114: This function will probably become quite big, so I think we could extract the broadcasting logic into a separate function

CarlTeapot (changes requested):

gateway/internal/hub/hub.go
ggela22: HandleWS is a per-connection goroutine, not the hub. hub only does register/unregister/getOrCreateRoom/discardIfStale, so single responsibility principle isn't broken here imo

the client sending directly to the room channel would mean either client validates the wire prefix or room does it, room should stay format agnostic.

if it gets complicated with snapshots and control messages in the future we can put the relay logic into its own file (e.g. connection_handler.go), but right now it's like 10 lines in a goroutine so not sure if it's worth it yet, but I can understand the logic in doing so

CarlTeapot: > the client sending directly to the room channel would mean either client validates the wire prefix or room does it, room should stay format agnostic.
> 
I think thats better than hub validating it. It makes less sense for hub to handle connections than client.go (which is literally for handling connections) 


> HandleWS is a per-connection goroutine, not the hub. hub only does register/unregister/getOrCreateRoom/discardIfStale, so single responsibility principle isn't broken here imo

it is a per-connection goroutine belonging to the hub instance, making it a part of the hub.

gateway/internal/protocol/protocol.go
ggela22: it's not for relays, this is for control messages documented in the comment block above it. seq is a server-assigned sequence number for future message ordering/dedup, not used yet but it was part of T04

gateway/internal/room/room.go
ggela22: good catch, will add drain

CarlTeapot: thanks gela 😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜😜

gateway/internal/room/room.go
ggela22: the window between ForceClose and Leave removing the client from the map is just goroutine scheduling, shouldn't be more than a couple milliseconds at most. the extra Send/ForceClose calls are don't do anything bc CloseSend already closed the channel and ForceClose does nothing when it sees it's already closing

keeping disconnected clients in the map with a state flag is session management and would probably make more sense to add with T16/T17, since a reconnecting client needs the snapshot implementation to catch up on missed ops anyway, otherwise they'd just be missing everything that happened while they were gone

gateway/internal/room/room.go
ggela22: I agree, will do

CarlTeapot (approved):

PR #67, opened 2026-05-02
add document state persistence and forking

### LINKED ISSUE #26: Implement Document state Persistence and Forking in Tauri

ggela22:
- added snapshot serialization to save complete document state using bitcode
- added .pcdoc file format with atomic writes to prevent corruption
- added save, load, list, and delete commands for document persistence
- added document forking

Closes #26


crdt-core/src/structs/block.rs
```diff
@@ -40,6 +40,29 @@ impl Block {
...
+    pub fn restore(
+        id: BlockId,
+        origin_left: Option<BlockId>,
+        origin_right: Option<BlockId>,
+        left: Option<BlockId>,
+        right: Option<BlockId>,
+        content: String,
+        is_deleted: bool,
+        len: u64,
+    ) -> Self {
```
CarlTeapot: we keep the len variable in the block to get the content length in O(1) time. but at restoring point, it really isn't a problem to calculate len with content.len. maybe we could get rid of the len variable to save disk space. what do you think? it is a very minor improvement but if's easy to do, why not?  not to mention if the file is corrupted somehow, they can be out of sync and can introduce problems

tauri-app/src-tauri/src/persistence/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+const FORMAT_VERSION: u8 = 1;
+const FILE_EXTENSION: &str = "pcdoc";
+
+#[derive(Debug)]
+pub enum PersistError {
+    Io(std::io::Error),
+    InvalidMagic,
+    UnsupportedFormat(u8),
+    Snapshot(SnapshotError),
+}
```
CarlTeapot, line 18: this file is named mod.rs. I am not sure if this is standard or not, if its standard you can ignore my comment but in other folders we always use mod.rs to just declare modules and the logic lives in separate files with better names. you could do the same here. your choice

crdt-core/src/snapshot.rs
```diff
@@ -0,0 +1,109 @@
...
+impl fmt::Display for SnapshotError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            SnapshotError::VersionMismatch { expected, got } => {
+                write!(
+                    f,
+                    "snapshot version mismatch: expected {expected}, got {got}"
+                )
+            }
+            SnapshotError::Decode(e) => write!(f, "snapshot decode failed: {e}"),
```
CarlTeapot, line 77: error!()  would be better because it fits in our logging system I think

crdt-core/src/snapshot.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    pub fn decode(bytes: &[u8]) -> Result<Self, SnapshotError> {
+        let snap: Snapshot = bitcode::decode(bytes).map_err(SnapshotError::Decode)?;
+        if snap.version != SNAPSHOT_VERSION {
+            return Err(SnapshotError::VersionMismatch {
+                expected: SNAPSHOT_VERSION,
+                got: snap.version,
+            });
+        }
```
CarlTeapot, line 103: This function is only called by the document::from_snapshot, which already does this check. You can remove this

tauri-app/src-tauri/src/persistence/commands.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+#[tauri::command]
+pub fn get_current_document_name(state: State<'_, AppState>) -> Result<Option<String>, String> {
+    Ok(state.current_document_name.lock().unwrap().clone())
+}
+
+#[tauri::command]
+pub fn save_text_file(path: String, content: String) -> Result<(), String> {
+    std::fs::write(&path, &content).map_err(|e| e.to_string())
```
CarlTeapot: just import this bruh

tauri-app/src-tauri/src/persistence/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    let payload = snap.encode();
+
+    let mut data = Vec::with_capacity(MAGIC.len() + 1 + payload.len());
+    data.extend_from_slice(MAGIC);
+    data.push(FORMAT_VERSION);
+    data.extend_from_slice(&payload);
+
+    let tmp = path.with_extension("pcdoc.tmp");
+    fs::write(&tmp, &data)?;
+    fs::rename(tmp, path)?;
```
CarlTeapot: This might not work on WIndows btw. I think it doesnt have atomic replace. if it fails you could remove file and then retry.  also, if it fails, the temp file is left on disk and isnt deleted, so we need to handle that too

tauri-app/src-tauri/src/persistence/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+        .path()
+        .app_data_dir()
+        .map_err(|e| PersistError::Io(std::io::Error::other(e.to_string())))?;
+    Ok(base.join("documents"))
+}
+
+fn doc_path(app: &AppHandle, name: &str) -> Result<PathBuf, PersistError> {
+    let dir = documents_dir(app)?;
+    Ok(dir.join(format!("{name}.{FILE_EXTENSION}")))
+}
```
CarlTeapot: I think windows has stricter rules for file naming, so we might need to add some checks

tauri-app/src-tauri/src/persistence/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    let snap = Snapshot::decode(&data[5..])?;
+    let doc = Document::from_snapshot(snap)?;
+    Ok(doc)
+}
+
+pub fn list_documents(app: &AppHandle) -> Result<Vec<DocumentMeta>, PersistError> {
+    let dir = documents_dir(app)?;
+    if !dir.exists() {
+        return Ok(Vec::new());
+    }
```
CarlTeapot: The exists() check is not atomic with the read_dir. If another process or the OS deletes the directory between the two calls you get an error rather than an empty list.


let iter = match fs::read_dir(&dir) {
Ok(it) => it,
Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(Vec::new()),
Err(e) => return Err(e.into()),
};

tauri-app/src-tauri/src/persistence/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+        docs.push(DocumentMeta {
+            name,
+            size_bytes: metadata.len(),
+            modified,
+        });
+    }
+
+    docs.sort_by_key(|d| std::cmp::Reverse(d.modified));
+    Ok(docs)
+}
```
CarlTeapot: If a file is placed in the folder manually (not via the app), name can contain characters that are forbidden by doc_path, causing load_named(name_from_meta) to fail or behave unexpectedly. Filter or normalize name before including in the resul

tauri-app/src-tauri/src/persistence/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    let mut data = Vec::with_capacity(MAGIC.len() + 1 + payload.len());
+    data.extend_from_slice(MAGIC);
+    data.push(FORMAT_VERSION);
+    data.extend_from_slice(&payload);
+
+    let tmp = path.with_extension("pcdoc.tmp");
+    fs::write(&tmp, &data)?;
+    fs::rename(tmp, path)?;
+    Ok(())
+}
```
CarlTeapot: This is almost identical to save_document. you could introduce a helper

CarlTeapot (changes requested):

crdt-core/src/snapshot.rs
ggela22: this is a fmt::Display impl, it formats the error as a string for the caller
logging error!() should happen where the error is handled, not inside Display

crdt-core/src/snapshot.rs
ggela22: I agree the check is redundant, but I'll keep it in decode() and remove it from from_snapshot() instead. better to have validation before it goes to from_snapshot imo

PR #70, opened 2026-05-03
PeerCode: integrate websocket to crdt

### LINKED ISSUE #43: Integrate Tauri Websocket receiver to CRDT
Tauri websocket receiver is currently a stub: it only logs any messages from the gateway.  Goal is to handle all types of messages from gateway and call corresponding CRDT functions. This task will also need a proper locking mechanism: the receiver is running in a separate thread and we need to synchronize it with IPC calls from frontend.

CarlTeapot:
* Added a feature to send raw serialized bytes to the gateway asynchronously using a channel
* Implemented a remote change listener to frontend
* Added buttons that lets you copy urls
* added remote_op_handler.rs file that validates remote operations and calls crdt api
* modified ws_receiver to call remote_op_handler
* modified the build command to use sccache for compiler caching and mold for fast dynamic linking for faster startup times
* added a script that automatically installs these tools if they don't exist on the host machine 
* fixed a small bug that failed the prod version building


tauri-app/src/remoteChangeListener.ts
```diff
@@ -0,0 +1,113 @@
...
+              forceMoveMarkers: true,
+            },
+          ]);
+
+          const count = ++eventCountRef.current;
+          const html =
+            `<span class="label">#${count}</span>` +
+            `<span class="op-insert">[remote-insert]</span> ` +
+            `offset=${change.position}  text=${JSON.stringify(change.content)}`;
+          setEventLog((prev) => [...prev, { id: count, html }]);
```
ChocoChichia: this could be a security issue, change if you want 👍 

The existing log panel renders log entries via dangerouslySetInnerHTML
(App.tsx:312). As I understadn now we have remote-change handlere that builds log html
strings out of change content arriving from websocket, thats untrusted content form any peer.

a peer typing  runs JS in every other peer's 
app via this code path.

tauri-app/src-tauri/src/crdt/local_op_handler.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    debug!(
+        "crdt insert succeeded: position={}, content_len={}",
+        position,
+        content.chars().count()
+    );
+
+    if let Some(wire_block) = wire_block_opt {
+        let frame = encode_op(&OpMessage::Insert(wire_block));
+        ws.send_raw(frame).await;
```
ChocoChichia, line 46: if I understand correctly flow is : lock document -> local_isnert mutates the crdt -> drop loclk ->
ws.send_raw(frame).await -> return ok(()) to the frontend
but send_raw (ws_state.rs:106-123) handles the disconnected case by warn!-ing and
returning. So if the websocket is in Disconnected state, the local doc gains 
the edit, the IPC handler reports success, the user types and sees the
character and the wire frame is dropped on the floor with no buffer and no retry.

to summarise the problem here :
The local
doc has been mutated, the IPC handler returns Ok(()) to the frontend, and 
the user sees the character on screen, but no peer ever receives it, and
there's no buffer to replay on reconnect.

I guess this can be a separate issue

CarlTeapot: I think this can be partially mitigated when we introduce permissions, we can block the read-only guests to modify their document state (we can disable typing in frontend and also add validations in tauri). We could also block typing for write permission guests if they disconnect with an option to exit the disconnected session. As for the Host, even if he is disconnected, he is still the source of truth so maybe its okay???

tauri-app/src-tauri/src/crdt/remote_op_handler.rs
```diff
@@ -0,0 +1,67 @@
...
+use crdt_core::{decode_op, OpMessage, RemoteChange};
+use log::{debug, error, warn};
+use tauri::{AppHandle, Emitter, Manager};
+
+use crate::state::appstate::AppState;
+use crate::ws_management::ws_types::RemoteChangeEvent;
+
+pub const REMOTE_CHANGE_EVENT: &str = "crdt://remote-change";
+
+pub fn handle_remote_binary(app: &AppHandle, bytes: &[u8]) {
```
ChocoChichia, line 22: This one function does three distinct jobs lets decompose

tauri-app/src-tauri/src/ws_management/ws_receiver.rs
```diff
@@ -21,6 +23,7 @@ pub async fn receive_loop(
             }
             Ok(Message::Binary(bytes)) => {
                 debug!("ws receiver binary message (bytes={})", bytes.len());
+                handle_remote_binary(&app, bytes.as_ref());
```
ChocoChichia: Synchronous CRDT integration runs on the async receiver task. while this handle remote binary ho is a sync fn.
it holds document mutex and integrates blocks. while this happens ( while the lock is held ) the work runs and the worker thread is blocked ( no other websocket frames are read, no other tasks scheduled on this thread make progress)

it is ok for one char inserts but for large paste and so on the receiver stalls.

ways out ( you could thin of better )  : actor pattern - single thread owns Document, receives OpMessages on a channel, returns Vec<RemoteChange> on a oneshot. Removes the mutex entirely. 
y.

PR #74, opened 2026-05-07
Implement Session leaving and ending 

### LINKED ISSUE #72: Add option to leave session without having to restart the app

CarlTeapot:
* decoupled process spawning from frontend emitting and room creation
* introduced a new gateway api layer that abstracts away gateway http communication
* moved host session commands and guest session commands in separate files
* moved process logger in a separate file
* Created a new Sidecar struct tu abstract some process spawning
* Created a feature to leave a session
* decoupled session ending and process killing
* implemented a feature to end session gracefully
* enhanced frontend
* implemented a flow to remove the room from go gateway
* modified make dev to use vite dev features
* updated dependency versions to fix vulnurabilities
* changed the debug logger tu rin in tokyo async task thread

### DIFF
--- a/gateway/internal/hub/hub.go
+++ b/gateway/internal/hub/hub.go
@@ -156,6 +156,30 @@ func dispatchFrame(rm *room.Room, sender *client.Client, raw []byte) {
 	}
 }
 
+func (h *Hub) HandleEndSession(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+	roomID := r.URL.Query().Get("room")
+	if roomID == "" {
+		http.Error(w, "missing ?room= parameter", http.StatusBadRequest)
+		return
+	}
+	h.mu.Lock()
+	rm, ok := h.rooms[roomID]
+	h.mu.Unlock()
+	if !ok {
+		slog.Warn("end-session: room not found", "room_id", roomID)
+		http.Error(w, "room not found", http.StatusNotFound)
+		return
+	}
+	frame := wire.EncodeControlFrame(wire.ControlSessionEnded)
+	rm.BroadcastAll(frame)
+	slog.Info("end-session: session-ended broadcast sent", "room_id", roomID)
+	w.WriteHeader(http.StatusOK)
+}
+
 func (h *Hub) HandleWS(w http.ResponseWriter, r *http.Request) {
 	roomID, clientID, ok := readWSParams(w, r)
 	if !ok {
--- a/tauri-app/src-tauri/src/gateway/gateway_api.rs
+++ b/tauri-app/src-tauri/src/gateway/gateway_api.rs
@@ -1,0 +1,70 @@
+use crate::gateway::types::{RoomResponse, GATEWAY_TIMEOUT};
+use log::{debug, warn};
+use url::Url;
+
+pub async fn create_room(port: u16) -> Result<String, String> {
+    debug!("fetching gateway room id via /rooms: port={port}");
+    tokio::time::timeout(GATEWAY_TIMEOUT, async {
+        reqwest::Client::new()
+            .post(format!("http://127.0.0.1:{port}/rooms"))
+            .send()
+            .await
+            .map_err(|e| format!("gateway /rooms: {e}"))?
+            .error_for_status()
+            .map_err(|e| format!("gateway /rooms: {e}"))?
+            .json::<RoomResponse>()
+            .await
+            .map(|r| r.room_id)
+            .map_err(|e| format!("gateway /rooms: {e}"))
+    })
+    .await
+    .map_err(|_| {
+        format!(
+            "gateway /rooms: timed out after {}s",
+            GATEWAY_TIMEOUT.as_secs()
+        )
+    })?
+}
+
+pub async fn destroy_room(local_room_url: String) -> Result<(), String> {
+    let end_session_url = end_session_url_from_room_url(&local_room_url)?;
+    tokio::time::timeout(GATEWAY_TIMEOUT, async {
+        reqwest::Client::new()
+            .post(&end_session_url)
+            .send()
+            .await
+            .map_err(|e| {
+                warn!("end-session HTTP request failed: {e}");
+                format!("Failed to notify gateway: {e}")
+            })?
+            .error_for_status()
+            .map_err(|e| {
+                warn!("end-session HTTP non-OK response: {e}");
+                format!("Gateway rejected end-session: {e}")
+            })?;
+        Ok::<(), String>(())
+    })
+    .await
+    .map_err(|_| {
+        format!(
+            "gateway /end-session: timed out after {}s",
+            GATEWAY_TIMEOUT.as_secs()
+        )
+    })??;
+    Ok(())
+}
+
+fn end_session_url_from_room_url(room_url: &str) -> Result<String, String> {
+    let parsed = Url::parse(room_url).map_err(|e| format!("invalid room URL: {e}"))?;
+    let room_id = parsed
+        .query_pairs()
+        .find(|(k, _)| k == "room")
+        .map(|(_, v)| v.into_owned())
+        .ok_or_else(|| "room URL is missing room query parameter".to_string())?;
+    let port = parsed
+        .port_or_known_default()
+        .ok_or_else(|| "room URL is missing a valid port".to_string())?;
+    Ok(format!(
+        "http://127.0.0.1:{port}/end-session?room={room_id}"
+    ))
+}
--- a/tauri-app/src-tauri/src/lib.rs
+++ b/tauri-app/src-tauri/src/lib.rs
@@ -1,6 +1,7 @@
 mod app_config;
 mod crdt;
 mod debug;
+mod gateway;
 mod persistence;
 mod processes;
 mod session;
@@ -12,10 +13,11 @@ use crate::app_config::identity;
 use crate::crdt::local_op_handler;
 #[cfg(debug_assertions)]
 use crate::debug::document_logger::spawn_linked_list_logger;
-use crate::state::appstate::AppState;
+use crate::gateway::gateway_api::destroy_room;
+use crate::state::appstate::{AppRole, AppState};
 use crate::state::ws_state::WsState;
 use crdt_core::types::ClientId;
-use log::{debug, info};
+use log::{debug, info, warn};
 use rand::random;
 use tauri::Manager;
 use tauri_plugin_log::{Target, TargetKind};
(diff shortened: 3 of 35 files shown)


tauri-app/src-tauri/src/session/host_commands.rs
```diff
@@ -0,0 +1,177 @@
...
+    drop(role);
+
+    let procs = state.processes.lock().unwrap();
+    let should_launch_processes = procs
+        .gateway
+        .as_ref()
+        .map(|s| s.status == SidecarStatus::Disabled)
+        .unwrap_or(true);
+    if !should_launch_processes {
+        return Err("Gateway sidecar is already running; refusing to launch a duplicate".into());
```
ChocoChichia, line 100: this err returns after line 82 is setting the role to starting.
host calls "rollback_starting_role" if prepare host session fails but in this case
user may permamently end up in starting and thats also a problem becausse 
your "can_initiate_session()" rejects every future host/join until the app is restarted

tauri-app/src-tauri/src/session/host_commands.rs
```diff
@@ -0,0 +1,177 @@
...
+                local_room_url,
+                ..
+            } => (room_id.clone(), local_room_url.clone()),
+            _ => {
+                warn!("end_session rejected: not a host");
+                return Err("Not currently hosting a session".into());
+            }
+        }
+    };
+    state.leave_session(&ws);
```
ChocoChichia, line 76: for guests leave session has : 
        if let AppRole::Guest { .. } = previous_role {
            ws.disconnect_nowait();
        }
for host at line 152 the local ws opened in connect() and 
is never explicitly torn down. what actually closes it today is the
gateway echoing our own session-ended controll frame back to us
(BroadcastAll doesn't exclude the sender), which is fragile.

you can call ws.disconnect.await here (after destroying the room or before idk)
so the host path does not depend on the echo.

tauri-app/src-tauri/src/session/host_commands.rs
```diff
@@ -0,0 +1,177 @@
...
+fn transition_to_host(app: &AppHandle, setup: &HostSessionSetup) {
+    let state = app.state::<AppState>();
+    let mut role = state.role.lock().unwrap();
+    *role = AppRole::Host {
+        room_id: setup.room_id.clone(),
+        lan_url: setup.lan_url.clone(),
+        public_url: setup.public_url.clone(),
+        local_room_url: setup.local_room_url.clone(),
+        public_room_url: setup.public_room_url.clone(),
+    };
```
ChocoChichia, line 134: transition_to_host overwrites role unconditionally
This unconditionally sets `*role = AppRole::Host { … }`
bebfore this pr we checked `matches!(*role, AppRole::Starting)` first and bailed if the role had
been changed concurrently
bro lost it during the refactor.

gateway/internal/hub/hub.go
```diff
@@ -156,6 +156,30 @@ func dispatchFrame(rm *room.Room, sender *client.Client, raw []byte) {
...
+	h.mu.Lock()
+	rm, ok := h.rooms[roomID]
+	h.mu.Unlock()
+	if !ok {
+		slog.Warn("end-session: room not found", "room_id", roomID)
+		http.Error(w, "room not found", http.StatusNotFound)
+		return
+	}
+	frame := wire.EncodeControlFrame(wire.ControlSessionEnded)
+	rm.BroadcastAll(frame)
```
ChocoChichia, line 178: Server broadcasts session-ended to the host itself

broadcastall calls getpeer(nil) so it includes the host that just POSTed /end-session
the hosts tauri app than nreceives its own session-ended controll frame lol,
back though its local ws, which 
leaks the hosts writer/processor tasks
and makes frontend show : the host has ended the session"


you can accept an optional  `?client_id=…` query param here and
pass it through to room.BroadcastAll(frame, exclude)The host's
tauri app already knows its own client_id (host_commands.rs:159),
just append it to the destroy URL in `gateway_api.rs::destroy_room`.

tauri-app/src/App.tsx
```diff
@@ -84,29 +337,56 @@ function AppContent({ username }: AppContentProps) {
...
         }),
       );
       unlisten.push(
         await listen<{ message: string }>("session://session-error", (e) => {
           setSessionStatus("error: " + e.payload.message);
+          setProcessesRunning({ gateway: "Disabled", tunnel: "Disabled" });
+        }),
+      );
+      unlisten.push(
+        await listen("session://session-ended", () => {
```
ChocoChichia, line 385: This banner fires for the host who just clicked End Session, because
the gateway echoes the control frame back (see hub.go:178). If you
fix that on the server, this listener still works for guests , but
worth gating on sessionStatus at firing time (`if (sessionStatus
=== "guest")`) as belt-and-suspenders, since this is a UX bug visible
to the user.

tauri-app/src-tauri/src/gateway/gateway_api.rs
```diff
@@ -0,0 +1,70 @@
...
+fn end_session_url_from_room_url(room_url: &str) -> Result<String, String> {
+    let parsed = Url::parse(room_url).map_err(|e| format!("invalid room URL: {e}"))?;
+    let room_id = parsed
+        .query_pairs()
+        .find(|(k, _)| k == "room")
+        .map(|(_, v)| v.into_owned())
+        .ok_or_else(|| "room URL is missing room query parameter".to_string())?;
+    let port = parsed
+        .port_or_known_default()
+        .ok_or_else(|| "room URL is missing a valid port".to_string())?;
```
ChocoChichia, line 66: port_known_or_default returns some(80) for ws:// urls without an exlicit port

( i googled and got : In Rust, port_or_known_default() is a method on the Url struct (typically from the url or reqwest crates) that returns the port number if it is explicitly in the URL, or the default port for the scheme (e.g., 80 for http) if it is not. )

so if local_room_url is ever reformatted to omit the port this will silently post to 
http://127.0.0.1:80/end-session and fail in confusing ways

axla sheidzleba DAKIDEBA idk

tauri-app/src-tauri/src/lib.rs
```diff
@@ -69,7 +70,9 @@ pub fn run() {
         .on_window_event(|window, event| {
             if let tauri::WindowEvent::Destroyed = event {
```
ChocoChichia, line 72: destroyed fires after the window is already gone so this only gets 
to do "kill_host_processes()", there is no chance to POST 
/end-session ( host ) or send a close frame ( guest ) 
giests connnected to a host who quit will see their ws drop with no 
session-ended signal, leaving UI inn  weird state.

you can un async cleanup (destroy_room or disconnect)
with a short timeout, then call `window.destroy()` or
`app.handle().exit(0)` on completion.

tauri-app/src-tauri/src/state/appstate.rs
```diff
@@ -14,13 +14,16 @@ pub struct AppState {
...
 pub enum AppRole {
     Undecided,
     Starting,
     Host {
         room_id: String,
         lan_url: Option<String>,
         public_url: Option<String>,
+        local_room_url: String,
+        public_room_url: Option<String>,
     },
```
ChocoChichia, line 27: optional

the role state machine is now mutated in 8+ call sites across
host_commands, gues_commans , joint_commands, gateway_process, ........

you could try moving transitions onto typed methods on appstate , like :  
  `try_mark_starting() -> Result<StartingGuard, AlreadyActive>`,
  `commit_host(StartingGuard, HostFields)`,
  `commit_guest(StartingGuard, GuestFields)`,
where `StartingGuard::drop` resets to `Undecided` unless committed.

ChocoChichia (changes requested):

tauri-app/src-tauri/src/ws_management/ws_receiver.rs
```diff
@@ -20,6 +26,18 @@ pub async fn receive_loop(
...
+                    match bytes.get(1).copied() {
+                        Some(CONTROL_SESSION_ENDED) => {
+                            info!("ws recv: session ended by host");
+                            let _ = app.emit(SESSION_ENDED, SessionEndedPayload {});
+                        }
+                        other => {
+                            warn!("ws recv: unknown control frame type={:?}; ignoring", other);
+                        }
+                    }
+                    break;
```
ggela22: shouldn't we move the break inside the session_ended check? right now it doesn't ignore unknown control frames

tauri-app/src-tauri/src/ws_management/ws_receiver.rs
```diff
@@ -2,16 +2,22 @@ use std::sync::{Arc, RwLock};
...
 use log::{debug, info, warn};
+use tauri::{AppHandle, Emitter};
 use tokio::sync::{mpsc, Mutex};
 use tokio_tungstenite::tungstenite::Message;
 
+use crate::session::session_types::{SessionEndedPayload, SESSION_ENDED};
 use crate::ws_management::ws_types::{Stream, WsConnection};
 
+const PREFIX_CONTROL: u8 = 0x02;
+const CONTROL_SESSION_ENDED: u8 = 0x01;
```
ggela22: the new control prefixes don't have a sync check. you should cover these in protocol_drift_tests.rs

also I think it'd be better to move them to src/wire

tauri-app/src/App.tsx
```diff
@@ -310,11 +666,159 @@ function AppContent({ username }: AppContentProps) {
...
+        {sessionStatus === "host" && (
+          <button
+            onClick={() => {
+              void invoke("end_session").then(() => {
+                setSessionStatus("idle");
+                setLanUrl(null);
+                setPublicUrl(null);
+              });
+            }}
+            style={{
```
ggela22, line 782: you use the same two button styles a lot of times, would be better to move them to a couple CSS classes instead of duplicating code

tauri-app/src/App.tsx
```diff
@@ -16,6 +29,156 @@ interface LogEntry {
...
 }
 
+// --- Save-before-session modal ---
+interface SavePromptProps {
+  onSaveAndContinue: (name: string) => Promise<void>;
+  onDiscardAndContinue: () => void;
+  onCancel: () => void;
+}
+
+function SaveBeforeSessionModal({
```
ggela22, line 39: this file is getting very bloated and it might be better to have SaveBeforeSessionModal & JoinModal be in their own files

ggela22 (changes requested):

tauri-app/src/App.tsx
CarlTeapot: fuck you

tauri-app/src/App.tsx
CarlTeapot: fuck you

tauri-app/src-tauri/src/ws_management/ws_receiver.rs
CarlTeapot: true

tauri-app/src-tauri/src/ws_management/ws_receiver.rs
CarlTeapot: true

tauri-app/src/App.tsx
ChocoChichia: bro is reviewing frontend lol

PR #76, opened 2026-05-12
implement snapshot distribution to new clients

### LINKED ISSUE #75: Implement Current state snapshot sending to a newly joined client

ggela22:
- host sends document snapshot to gateway on connect
- gateway stores snapshot + buffers all ops afterwards, replays both to new joiners
- guest decodes snapshot, forks document with own client_id
- host re-sends fresh snapshot every 100 ops to keep buffer bounded
- snapshot wire encode/decode added to crdt-core

### DIFF
--- a/gateway/internal/client/client.go
+++ b/gateway/internal/client/client.go
@@ -47,6 +47,11 @@ func (c *Client) Send(data []byte) (ok bool) {
 	}
 }
 
+// for testing only
+func (c *Client) SendChan() <-chan []byte {
+	return c.send
+}
+
 func (c *Client) ForceClose() {
 	if c.conn == nil {
 		slog.Debug("force-close skipped: client connection is nil", "room_id", c.RoomID, "client_id", c.ID)
--- a/gateway/internal/room/room.go
+++ b/gateway/internal/room/room.go
@@ -168,13 +168,13 @@ func (r *Room) BroadcastAll(data []byte) {
 
 func (r *Room) getPeers(exclude *client.Client) []*client.Client {
 	r.mu.Lock()
+	defer r.mu.Unlock()
 	targets := make([]*client.Client, 0, len(r.clients))
 	for c := range r.clients {
 		if exclude == nil || c != exclude {
 			targets = append(targets, c)
 		}
 	}
-	r.mu.Unlock()
 	return targets
 }
 
--- a/gateway/internal/room/room.go
+++ b/gateway/internal/room/room.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 
 	"gateway/internal/client"
+	"gateway/internal/wire"
 )
 
 const opsBufferSize = 256
@@ -20,6 +21,9 @@ type Room struct {
 	mu      sync.Mutex
 	clients map[*client.Client]struct{}
 	closed  bool
+
+	latestSnapshot []byte
+	opsLog         [][]byte
 
 	ops  chan BroadcastMsg
 	done chan struct{}
@@ -77,6 +81,53 @@ func (r *Room) Size() int {
 	return len(r.clients)
 }
 
+func (r *Room) StoreSnapshot(data []byte) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.latestSnapshot = make([]byte, len(data))
+	copy(r.latestSnapshot, data)
+	r.opsLog = nil
+	slog.Info("snapshot stored", "room_id", r.ID, "bytes", len(data))
+}
+
+func (r *Room) AppendOp(data []byte) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.latestSnapshot == nil {
+		return
+	}
+	cp := make([]byte, len(data))
+	copy(cp, data)
+	r.opsLog = append(r.opsLog, cp)
+}
+
+func (r *Room) ReplayTo(c *client.Client) bool {
+	r.mu.Lock()
+	snap := r.latestSnapshot
+	ops := make([][]byte, len(r.opsLog))
+	copy(ops, r.opsLog)
+	r.mu.Unlock()
+
+	if snap == nil {
+		slog.Debug("replay skipped: no snapshot available", "room_id", r.ID, "client_id", c.ID)
+		return false
+	}
+
+	if !c.Send(snap) {
+		slog.Warn("replay: failed to send snapshot to joiner", "room_id", r.ID, "client_id", c.ID)
+		return false
+	}
+	slog.Info("replay: snapshot sent to joiner", "room_id", r.ID, "client_id", c.ID, "snapshot_bytes", len(snap), "buffered_ops", len(ops))
+
+	for i, op := range ops {
+		if !c.Send(op) {
+			slog.Warn("replay: failed to send buffered op to joiner", "room_id", r.ID, "client_id", c.ID, "op_index", i)
+			break
+		}
+	}
+	return true
+}
+
 func (r *Room) Run() {
 	slog.Info("room loop started", "room_id", r.ID)
 	for {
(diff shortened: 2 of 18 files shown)


tauri-app/src/App.tsx
```diff
@@ -123,6 +123,48 @@ function AppContent({ username }: AppContentProps) {
...
+
+    return () => {
+      cancelled = true;
+      if (unlisten) {
+        unlisten();
+        unlisten = null;
+      }
+    };
+  }, [editorRef, isApplyingRemote, eventCountRef, setEventLog]);
+
```
CarlTeapot, line 417: Please move this to a separate file if possible

tauri-app/src-tauri/src/session/command.rs
```diff
@@ -236,5 +236,17 @@ async fn connect(app: AppHandle, port: u16, room_id: String) {
...
+            let doc = state.document.lock().unwrap();
+            let snap = doc.to_snapshot();
+            crdt_core::encode_snapshot(&snap)
+        };
+        ws.send_raw(snapshot_frame).await;
+        app.state::<AppState>()
+            .ops_since_snapshot
+            .store(0, std::sync::atomic::Ordering::Relaxed);
+        info!("host sent initial document snapshot to gateway");
     }
```
CarlTeapot: You could move this to a separate function for single responsibility

gateway/internal/client/client.go
```diff
@@ -47,6 +47,11 @@ func (c *Client) Send(data []byte) (ok bool) {
 	}
 }
 
+// for testing only
+func (c *Client) SendChan() <-chan []byte {
+	return c.send
+}
```
CarlTeapot, line 53: are you sure that this is a good solution? maybe you could make the channel injectable like when clock instances are injected in services to make it easier to mock time in tests

gateway/internal/room/room.go
```diff
@@ -77,6 +81,53 @@ func (r *Room) Size() int {
...
+	r.mu.Lock()
+	snap := r.latestSnapshot
+	ops := make([][]byte, len(r.opsLog))
+	copy(ops, r.opsLog)
+	r.mu.Unlock()
+
+	if snap == nil {
+		slog.Debug("replay skipped: no snapshot available", "room_id", r.ID, "client_id", c.ID)
+		return false
+	}
```
CarlTeapot, line 114: if the snap is null, is the operations log empty too? 
if not, maybe it would be more optimal to do this check before the copy()

gateway/internal/room/room.go
```diff
@@ -87,20 +138,18 @@ func (r *Room) Run() {
 			slog.Info("room loop stopped", "room_id", r.ID)
 			return
 		case msg := <-r.ops:
-			r.broadcast(msg)
+			if wire.IsSnapshotFrame(msg.Data) {
+				r.StoreSnapshot(msg.Data)
+			} else {
+				r.AppendOp(msg.Data)
+				r.broadcast(msg)
+			}
```
CarlTeapot, line 146: The source of truth in our version of crdt is the host. so maybe we could add a check that only accepts snapshot frames from the host himself

gateway/internal/room/room.go
```diff
@@ -109,9 +158,30 @@ func (r *Room) broadcast(msg BroadcastMsg) {
...
+func (r *Room) getPeers(exclude *client.Client) []*client.Client {
+	r.mu.Lock()
+	targets := make([]*client.Client, 0, len(r.clients))
+	for c := range r.clients {
+		if exclude == nil || c != exclude {
+			targets = append(targets, c)
+		}
+	}
+	r.mu.Unlock()
+	return targets
```
CarlTeapot: defer bro

gateway/internal/room/room.go
```diff
@@ -109,9 +158,30 @@ func (r *Room) broadcast(msg BroadcastMsg) {
...
+
+func (r *Room) sendToPeers(targets []*client.Client, data []byte, slowClientLogMsg string) {
 	for _, c := range targets {
-		if !c.Send(msg.Data) {
-			slog.Warn("disconnecting slow client", "room_id", r.ID, "client_id", c.ID)
+		if !c.Send(data) {
+			slog.Warn(slowClientLogMsg, "room_id", r.ID, "client_id", c.ID)
 			c.ForceClose()
 		}
 	}
```
CarlTeapot, line 187: We could optimize this by making individual sending a goroutine and sending them in parallel

tauri-app/src-tauri/src/crdt/local_op_handler.rs
```diff
@@ -83,11 +86,31 @@ pub async fn delete(
...
+        maybe_send_snapshot(&state, &ws).await;
     }
 
     Ok(())
 }
 
+async fn maybe_send_snapshot(state: &State<'_, AppState>, ws: &State<'_, WsState>) {
+    if !matches!(
+        *state.role.lock().unwrap(),
+        crate::state::appstate::AppRole::Host { .. }
```
CarlTeapot: just import bro

tauri-app/src-tauri/src/crdt/local_op_handler.rs
```diff
@@ -83,11 +86,31 @@ pub async fn delete(
...
+    let count = state.ops_since_snapshot.fetch_add(1, Ordering::Relaxed) + 1;
+    if count >= SNAPSHOT_REFRESH_INTERVAL {
+        state.ops_since_snapshot.store(0, Ordering::Relaxed);
+        let snapshot_frame = {
+            let doc = state.document.lock().unwrap();
+            crdt_core::encode_snapshot(&doc.to_snapshot())
+        };
+        ws.send_raw(snapshot_frame).await;
+        info!("periodic snapshot sent (after {} ops)", count);
+    }
```
CarlTeapot, line 102: you could inverse the if statement to reduce nesting

tauri-app/src-tauri/src/crdt/remote_op_handler.rs
```diff
@@ -1,15 +1,16 @@
...
 use log::{debug, error, info, warn};
 use tauri::{AppHandle, Emitter, Manager};
 use tokio::sync::mpsc::UnboundedReceiver;
 
 use crate::state::appstate::AppState;
-use crate::ws_management::ws_types::RemoteChangeEvent;
+use crate::ws_management::ws_types::{RemoteChangeEvent, SnapshotAppliedEvent};
 
 pub const REMOTE_CHANGE_EVENT: &str = "crdt://remote-change";
+pub const SNAPSHOT_APPLIED_EVENT: &str = "crdt://snapshot-applied";
```
CarlTeapot: both of these constants can be private i think

CarlTeapot (changes requested):

gateway/internal/room/room.go
CarlTeapot: If you decide to do this, you could do the same in tauri so that guest client doesn't send these frames for no reason

gateway/internal/room/room.go
ggela22: yes, opsLog is always empty when snap == nil, bc AppendOp returns early when latestSnapshot == nil

gateway/internal/room/room.go
ggela22: we already have a role check in maybe_send_snapshot, I don't know if it'd be a good design choice to have the gateway track who the host is

gateway/internal/room/room.go
ggela22: ოკ

gateway/internal/room/room.go
ggela22: c.send is a channel write, it's O(1), I don't see the benefit of adding goroutines

tauri-app/src-tauri/src/crdt/local_op_handler.rs
ggela22: kargi chonker

PR #85, opened 2026-05-13
Implement authentication in client -> gateway communication

### LINKED ISSUE #57: Introduce Authentication to improve security in client -> gateway communication
1. Room squatting / eavesdropping Room IDs are 8 hex characters (4 bytes = ~4 billion combinations). Anyone who can guess or brute-force a room ID can silently join and receive every CRDT op broadcast to that room. There is no invite, no password, no token.

2. Identity spoofing Any client can connect with client_id=alice and the gateway will happily register them as Alice. There is nothing preventing two connections with the same client_id in the same room — the room's clients map uses *client.Client pointer keys, so duplicates would silently coexist, but all ops would appear to come from "alice".

3. Unbounded joining There is no rate limiting, no connection cap per room, and no per-IP throttling. Anyone can open thousands of connections to a room.

CarlTeapot:
* Added a secure token generation that is passed to the gateway as an env. var. at startup
  * added a filter to the gateway that invalidates any http request that doesnt contain the token in a header
  * modified gateway api to send the token on every http call
  * added a check in gateway to prevent peers with matching client_id from joining the server to prevent brute forcing
  * made room_id-s longer to prevent brute force attacks
  * Added rate limiting to websocket connection to avoid brute force attacks

### DIFF
--- a/gateway/cmd/server/ws_limit.go
+++ b/gateway/cmd/server/ws_limit.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"log/slog"
+	"net"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/didip/tollbooth/v8"
+	"github.com/didip/tollbooth/v8/libstring"
+	"github.com/didip/tollbooth/v8/limiter"
+)
+
+func clientIPForWSRateLimit(r *http.Request) string {
+	if s := strings.TrimSpace(r.Header.Get("CF-Connecting-IP")); s != "" {
+		return libstring.CanonicalizeIP(s)
+	}
+	if s := strings.TrimSpace(r.Header.Get("X-Forwarded-For")); s != "" {
+		parts := strings.Split(s, ",")
+		for i := range parts {
+			parts[i] = strings.TrimSpace(parts[i])
+		}
+		if len(parts) > 0 && parts[0] != "" {
+			return libstring.CanonicalizeIP(parts[0])
+		}
+	}
+	host, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		return libstring.CanonicalizeIP(strings.TrimSpace(r.RemoteAddr))
+	}
+	return libstring.CanonicalizeIP(host)
+}
+
+func newWSRateLimiter(rpm int) *limiter.Limiter {
+	max := float64(rpm) / 60.0
+	return tollbooth.NewLimiter(max, nil)
+}
+
+func wsRateLimitMiddleware(lmt *limiter.Limiter, next http.HandlerFunc) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ip := clientIPForWSRateLimit(r)
+		if ip == "" {
+			ip = "unknown"
+		}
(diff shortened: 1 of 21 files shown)


gateway/internal/hub/hub.go
```diff
@@ -27,12 +28,12 @@ func New() *Hub {
 	return &Hub{rooms: make(map[string]*room.Room)}
 }
 
-func (h *Hub) newRoomID() string {
+func newRoomID() (string, error) {
 	var b [4]byte
```
ggela22, line 33: you didn't actually make roomids longer

tauri-app/src-tauri/src/gateway/auth_token_generator.rs
```diff
@@ -0,0 +1,12 @@
+use rand::rngs::OsRng;
+use rand::RngCore;
+
+pub fn generate_gateway_token() -> String {
+    let mut bytes = [0u8; 32];
+    OsRng.fill_bytes(&mut bytes);
+    bytes.iter().fold(String::with_capacity(64), |mut s, b| {
```
ggela22: just use hex::encode bro

gateway/internal/hub/hub.go
```diff
@@ -73,24 +79,37 @@ func (h *Hub) discardIfStale(id string, r *room.Room) {
 	}
 }
 
-func (h *Hub) register(c *client.Client) *room.Room {
+func (h *Hub) register(c *client.Client) (*room.Room, error) {
```
ggela22, line 83: maybe we should have a max retry count here?

ggela22 (changes requested):

gateway/internal/hub/hub.go
CarlTeapot: @ggela22 No need, IP rate limiting already solves the brute forcing

gateway/cmd/server/ws_limit.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/didip/tollbooth/v8/libstring"
+	"github.com/didip/tollbooth/v8/limiter"
+)
+
+func clientIPForWSRateLimit(r *http.Request) string {
+	if s := strings.TrimSpace(r.Header.Get("CF-Connecting-IP")); s != "" {
+		return libstring.CanonicalizeIP(s)
+	}
+	if s := strings.TrimSpace(r.Header.Get("X-Forwarded-For")); s != "" {
+		parts := strings.Split(s, ",")
```
ChocoChichia, line 21: The rate limiter tracks "how many times has this IP connected?"
but the IP comes from a header that anyone can fake.
X-Forwarded-For is just a line in the HTTP request. Like the room ID
query param, like any other field

so x-forwarded-for 4.4.4.4 could also work I think
idk ideal solution to that 

do not trust the header like that if you planned to save the server from attacks then add safe check also.
you can use env variables to check if configured

CarlTeapot: Ip spoofing can't happen because websocket connection needs a tcp handshake when the connection is established. if the attacker spoofs his IP, the handshake will never complete (the ack packet will go to the spoofed ip, not to the attack).

PR #89, opened 2026-05-14
add permission management

### LINKED ISSUE #27: Implement Permission management system in the gateway server
Right now, there is no permission management system in the gateway server. its just a stupid, dumb, chopped relay. We need to implement a way to integrate permission control layer to the existing system and display the current state of the entire room to all clients. default permission level should be configurable by host when creating the session

Implementation idea:
frontend has a panel that displays all current peers and their permissions. this panel can be opened/closed. for the host it also has a button that can give/revoke write access to the user. 

Backend flow:
* Host gives write access to guest 
*  frontend calls tauri via ipc 
* dedicated tauri handler sends the permission change event to the gateway server 
* gateway server updates its in-memory map of user permissions  relays the change to every other client belonging to the same room 
* websocket receivers in other clients receive the event and update the frontend 
*  the user impacted by the change finally gets the permission to write code

ggela22:
guests join as read-only by default, host can grant/revoke write access per peer in real-time.

- new control frame types for permission change/room state
- per-room permissions map, host identity verification via auth token, room state broadcasts on join/leave/permission change
- set_peer_permission IPC command, "can_write" guard on insert/delete, auth token lifecycle
- PeersPanel component with toggle switches for permissions
- updated Go tests to work with new permission functionality

### DIFF
--- a/gateway/internal/client/client.go
+++ b/gateway/internal/client/client.go
@@ -10,20 +10,24 @@ import (
 const sendBufferSize = 256
 
 type Client struct {
-	ID     string
-	RoomID string
-	conn   *websocket.Conn
-	send   chan []byte
+	ID       string
+	RoomID   string
+	Username string
+	IsHost   bool
+	conn     *websocket.Conn
+	send     chan []byte
 }
 
-func New(id, roomID string, conn *websocket.Conn) *Client {
+func New(id, roomID, username string, isHost bool, conn *websocket.Conn) *Client {
 	c := &Client{
-		ID:     id,
-		RoomID: roomID,
-		conn:   conn,
-		send:   make(chan []byte, sendBufferSize),
+		ID:       id,
+		RoomID:   roomID,
+		Username: username,
+		IsHost:   isHost,
+		conn:     conn,
+		send:     make(chan []byte, sendBufferSize),
 	}
-	slog.Info("client created", "room_id", roomID, "client_id", id, "send_buffer_size", sendBufferSize)
+	slog.Info("client created", "room_id", roomID, "client_id", id, "username", username, "is_host", isHost, "send_buffer_size", sendBufferSize)
 	return c
 }
 
--- a/gateway/internal/room/room.go
+++ b/gateway/internal/room/room.go
@@ -1,11 +1,17 @@
 package room
 
 import (
+	"errors"
 	"log/slog"
 	"sync"
 
 	"gateway/internal/client"
 	"gateway/internal/wire"
+)
+
+var (
+	ErrRoomClosed        = errors.New("room closed")
+	ErrDuplicateClientID = errors.New("duplicate client_id")
 )
 
 const opsBufferSize = 256
@@ -22,8 +28,10 @@ type Room struct {
 	clients map[*client.Client]struct{}
 	closed  bool
 
-	latestSnapshot []byte
-	opsLog         [][]byte
+	latestSnapshot  []byte
+	opsLog          [][]byte
+	permissions     map[string]bool
+	defaultCanWrite bool
 
 	ops  chan BroadcastMsg
 	done chan struct{}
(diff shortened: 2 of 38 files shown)


gateway/internal/client/client.go
```diff
@@ -10,20 +10,24 @@ import (
...
 type Client struct {
-	ID     string
-	RoomID string
-	conn   *websocket.Conn
-	send   chan []byte
+	ID       string
+	RoomID   string
+	Username string
+	IsHost   bool
+	conn     *websocket.Conn
```
CarlTeapot, line 17: We could have a Role enum: 
type Role int

const (
    Host Role = iota
    Guest
)

gateway/internal/room/room.go
```diff
@@ -28,19 +28,23 @@ type Room struct {
...
 func New(id string) *Room {
 	return &Room{
-		ID:      id,
-		clients: make(map[*client.Client]struct{}),
-		ops:     make(chan BroadcastMsg, opsBufferSize),
-		done:    make(chan struct{}),
+		ID:              id,
+		clients:         make(map[*client.Client]struct{}),
+		permissions:     make(map[string]bool),
+		defaultCanWrite: false,
```
CarlTeapot, line 45: why are we hardcoding this? this should be chosen by the user when hosting the session in frontend

gateway/internal/room/room.go
```diff
@@ -28,19 +28,23 @@ type Room struct {
 	clients map[*client.Client]struct{}
 	closed  bool
 
-	latestSnapshot []byte
-	opsLog         [][]byte
+	latestSnapshot  []byte
+	opsLog          [][]byte
+	permissions     map[string]bool
```
CarlTeapot, line 33: why the separate map? it can be part of the clients map. client map can have client id as a key and client struct as a value. the client struct should contain the permission level

gateway/internal/room/room.go
```diff
@@ -160,6 +176,80 @@ func (r *Room) Run() {
...
+	for c := range r.clients {
+		peers = append(peers, PeerInfo{
+			ClientID: c.ID,
+			Username: c.Username,
+			IsHost:   c.IsHost,
+			CanWrite: r.permissions[c.ID],
+		})
+	}
+	return RoomStatePayload{Peers: peers}
+}
```
CarlTeapot, line 235: We could maybe copy the clients map and then unlock the mutex, not sure

gateway/internal/room/room.go
```diff
@@ -160,6 +176,80 @@ func (r *Room) Run() {
...
+	if err != nil {
+		slog.Error("failed to encode room state", "room_id", r.ID, "error", err)
+		return
+	}
+	targets := r.getPeers(nil)
+	for _, c := range targets {
+		if !c.Send(frame) {
+			slog.Warn("failed to send room state to client", "room_id", r.ID, "client_id", c.ID)
+		}
+	}
```
CarlTeapot, line 249: the broadcasting could happen in parallel

gateway/internal/room/room.go
```diff
@@ -160,6 +176,80 @@ func (r *Room) Run() {
...
+}
+
+func (r *Room) SetPermission(clientID string, canWrite bool) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if _, exists := r.permissions[clientID]; exists {
+		r.permissions[clientID] = canWrite
+	}
+}
```
CarlTeapot, line 209: why are we locking the entire room mutex to change a permission level of 1 client?

gateway/internal/room/room.go
```diff
@@ -160,6 +176,80 @@ func (r *Room) Run() {
...
+type PeerInfo struct {
+	ClientID string `json:"client_id"`
+	Username string `json:"username"`
+	IsHost   bool   `json:"is_host"`
+	CanWrite bool   `json:"can_write"`
+}
+
+type RoomStatePayload struct {
+	Peers []PeerInfo `json:"peers"`
+}
```
CarlTeapot, line 220: These struct shouldn't be here. the structs should either be in a separate file or be at the top of this file

gateway/internal/room/room.go
```diff
@@ -160,6 +176,80 @@ func (r *Room) Run() {
...
+			TargetClientID string `json:"target_client_id"`
+			CanWrite       bool   `json:"can_write"`
+		}
+		if err := wire.DecodeControlJSON(msg.Data, &change); err != nil {
+			slog.Warn("permission change decode failed", "room_id", r.ID, "error", err)
+			return
+		}
+		r.SetPermission(change.TargetClientID, change.CanWrite)
+		r.BroadcastRoomState()
+		slog.Info("permission updated", "room_id", r.ID, "target", change.TargetClientID, "can_write", change.CanWrite)
```
CarlTeapot, line 197: Not that important, but we could just send a permission update frame containing only the client whose permission changed. resending the entire state feels wasteful, but it is a very minor improvement, so your choice

tauri-app/src-tauri/src/app_config/identity.rs
```diff
@@ -68,6 +68,13 @@ pub fn get_identity(app: AppHandle) -> Result<IdentityDto, String> {
...
+pub fn read_username(app: &AppHandle) -> String {
+    load_raw(app)
+        .ok()
+        .and_then(|id| id.username)
+        .unwrap_or_default()
+}
+
 #[tauri::command]
 pub fn set_username(app: AppHandle, username: String) -> Result<(), String> {
     info!(
```
CarlTeapot, line 80: Not a part of this PR, but could you please move the tauri command functions in a separate commands.rs file? I think it's a good way to separate responsibilities and makes it similar to isolating endpoint and controller implementations in web apps

tauri-app/src-tauri/src/state/appstate.rs
```diff
@@ -11,6 +11,7 @@ pub struct AppState {
     pub processes: Mutex<HostProcesses>,
     pub current_document_name: Mutex<Option<String>>,
     pub ops_since_snapshot: AtomicU32,
+    pub can_write: AtomicBool,
     #[cfg(debug_assertions)]
```
CarlTeapot, line 15: What do you think about making it an enum and making that enum a part of the app role?

tauri-app/src-tauri/src/ws_management/ws_receiver.rs
```diff
@@ -68,3 +83,35 @@ pub async fn receive_loop(
...
+    if let Some(me) = room_state
+        .peers
+        .iter()
+        .find(|p| p.client_id == local_client_id)
+    {
+        state.can_write.store(me.can_write, Ordering::Relaxed);
+        let _ = app.emit(CAN_WRITE_EVENT, me.can_write);
+        debug!("local write permission updated: can_write={}", me.can_write);
+    }
+}
```
CarlTeapot, line 117: This shouldn't be decided by the websocket receiver. This is the responsibility of app state or something else

CarlTeapot (changes requested):

PR #99, opened 2026-05-18
Bugfix/forntend sync

ChocoChichia:
Sampled `baseSeq` at the moment Monaco fires the change event, passes it through
`enqueueOp` as a plain value.

### DIFF
--- a/tauri-app/src/App.tsx
+++ b/tauri-app/src/App.tsx
@@ -3,6 +3,7 @@ import {
   useRef,
   useEffect,
   useCallback,
+  useMemo,
   type FormEvent,
 } from "react";
 import type { editor } from "monaco-editor";
@@ -11,6 +12,7 @@ import { invoke } from "@tauri-apps/api/core";
 import { listen } from "@tauri-apps/api/event";
 import { useRemoteChangeListener } from "./remoteChangeListener";
 import { useSnapshotListener } from "./snapshotListener";
+import { createEnqueueOp, createIpcSenders } from "./opQueue";
 import {
   UsernameGate,
   overlayStyle,
@@ -220,11 +222,11 @@ function AppContent({ username }: AppContentProps) {
   const lastAppliedSeqRef = useRef(0);
   const opChainRef = useRef<Promise<unknown>>(Promise.resolve());
 
-  const enqueueOp = useCallback(<T,>(task: () => Promise<T>): Promise<T> => {
-    const next = opChainRef.current.then(task, task);
-    opChainRef.current = next.catch(() => undefined);
-    return next;
-  }, []);
+  const enqueueOp = useMemo(() => createEnqueueOp(opChainRef), []);
+  const { sendInsert, sendDelete } = useMemo(
+    () => createIpcSenders(enqueueOp),
+    [enqueueOp],
+  );
 
   const handleDocumentLoaded = useCallback((text: string, name: string) => {
     const ed = editorRef.current;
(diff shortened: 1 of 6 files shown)


tauri-app/src/opQueue.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function createEnqueueOp(
+  opChainRef: RefObject<Promise<unknown>>,
+): EnqueueOp {
+  return <T>(task: () => Promise<T>): Promise<T> => {
+    const next = opChainRef.current.then(task, task);
+    opChainRef.current = next.catch(() => undefined);
+    return next;
+  };
+}
```
CarlTeapot, line 19: Claudem mitxra ro: 
RefObject<T> in React's type definitions declares current as readonly. Writing opChainRef.current = ... is a type error under strict mode.  useRef<Promise<unknown>>(Promise.resolve()) returns MutableRefObject<Promise<unknown>>, which is what should be declared in the parameter. This may not cause a runtime issue (the readonly is only in types), but it breaks strict TypeScript checking and will confuse future readers. Change the parameter type to MutableRefObject<Promise<unknown>>.

tauri-app/src/App.tsx
```diff
@@ -529,10 +514,10 @@ function AppContent({ username }: AppContentProps) {
...
             try {
               if (deleteLen > 0) {
-                await sendDelete(offset, deleteLen);
+                await sendDelete(offset, deleteLen, baseSeq);
               }
               if (insertText.length > 0) {
-                await sendInsert(offset, insertText);
+                await sendInsert(offset, insertText, baseSeq);
               }
             } catch (error) {
```
CarlTeapot, line 522: baseSeq is captured once per Monaco event and used for every entry in event.changes. That's correct — Monaco gives all offsets relative to the pre-batch document, which is exactly the state baseSeq describes. The problem is on the backend side: OpLog::transform only adjusts positions for remote ops; it has no record of local ops (they're never pushed to the log). So when the loop awaits iteration 1's insert and then fires iteration 2, the backend has already shifted the document by the first insert but transform(offset₂, baseSeq) produces the same result as if it hadn't.

possible fix:

const baseSeq = lastAppliedSeqRef.current;
let localDelta = 0;
for (const change of event.changes) {
const offset = change.rangeOffset + localDelta;
if (deleteLen > 0) {
await sendDelete(offset, deleteLen, baseSeq);
localDelta -= deleteLen;
}
if (insertText.length > 0) {
await sendInsert(offset, insertText, baseSeq);
localDelta += insertText.length;
}
}

CarlTeapot (changes requested):

tauri-app/src/App.tsx
ChocoChichia: Monaco's API contract: event.changes is sorted in descending rangeOffset
order. The change at offset 8 fires FIRST, the change at offset 0 fires
SECOND.

Walk through with that order:
1. delete+insert at offset 8: "foo bar foo" → "foo bar wally". Position
0 still points at the same f. 
2. delete+insert at offset 0: "foo bar wally" → "wally bar wally".  ( nice )

The high-offset op doesn't affect low-offset positions, so the lower op
never needs transform compensation.


your fix would not work 
Their localDelta patch assumes forward order. Applied to Monaco's actual
reverse order:
- First iteration at offset=8, replace foo→wally: localDelta += 2
- Second iteration at offset=0: sends 0 + 2 = 2 → wrong! Should send 0.

PR #100, opened 2026-05-18
Implement Atomic local replace function:

### LINKED ISSUE #83: Send Monaco `replace` as a single atomic IPC
### Problem

`App.tsx:507-513` splits a Monaco `replace` event (paste, autocomplete,
find-and-replace) into a `sendDelete` then a `sendInsert`, **awaiting between
them**. A remote op can land on the actor between the two halves, shifting
the CRDT state so the second half operates on the wrong coordinates. Even
after Issue 1 is fixed, the two halves still drift apart because the actor
processes them as independent ops.

### Fix

Add an atomic `replace` DocOp on the Rust side; route Monaco replace events
through it from JS.

```rust
// state/document/types.rs - extend DocOp
LocalReplace {
    position: u64,
    delete_length: u64,
    content: String,
    base_seq: u64,
    reply: oneshot::Sender<Result<(), String>>,
},
```

```rust
// state/document/handlers/local.rs - new handler
pub fn handle_local_replace(
    state: &mut DocState,
    position: u64,
    delete_length: u64,
    content: &str,
    base_seq: u64,
) -> Result<(WireBlock, DeleteSet), String> {
    let transformed = state.op_log.transform(position, base_seq);
    let ds = delete_with_clamp(state, transformed, delete_length)?;
    let wire = insert_with_clamp(state, transformed, content)?
        .ok_or_else(|| "empty replace content".to_string())?;
    Ok((wire, ds))
}
```

Register one `#[tauri::command] pub async fn replace(...)` mirroring the
existing `insert` / `delete` shape, then collapse the JS handler:

```ts
// App.tsx — one IPC for replace
if (deleteLen > 0 && insertText.length > 0) {
  await enqueueOp(() =>
    invoke("replace", {
      position: offset,
      deleteLength: deleteLen,
      content: insertText,
      baseSeq,
    }),
  );
} else if (deleteLen > 0) {
  await sendDelete(offset, deleteLen, baseSeq);
} else if (insertText.length > 0) {
  await sendInsert(offset, insertText, baseSeq);
}
```

CarlTeapot:
* Implemented a new document operation that atomically replaces a text with different text and making sure that no remote operations happen between them

### DIFF
--- a/tauri-app/src/App.tsx
+++ b/tauri-app/src/App.tsx
@@ -231,7 +231,7 @@ function AppContent({ username }: AppContentProps) {
   const opChainRef = useRef<Promise<unknown>>(Promise.resolve());
 
   const enqueueOp = useMemo(() => createEnqueueOp(opChainRef), []);
-  const { sendInsert, sendDelete } = useMemo(
+  const { sendInsert, sendDelete, sendReplace } = useMemo(
     () => createIpcSenders(enqueueOp),
     [enqueueOp],
   );
@@ -513,10 +513,11 @@ function AppContent({ username }: AppContentProps) {
             });
 
             try {
-              if (deleteLen > 0) {
+              if (deleteLen > 0 && insertText.length > 0) {
+                await sendReplace(offset, deleteLen, insertText, baseSeq);
+              } else if (deleteLen > 0) {
                 await sendDelete(offset, deleteLen, baseSeq);
-              }
-              if (insertText.length > 0) {
+              } else {
                 await sendInsert(offset, insertText, baseSeq);
               }
             } catch (error) {
--- a/tauri-app/src-tauri/src/lib.rs
+++ b/tauri-app/src-tauri/src/lib.rs
@@ -106,6 +106,7 @@ pub fn run() {
         .invoke_handler(tauri::generate_handler![
             doc_commands::insert,
             doc_commands::delete,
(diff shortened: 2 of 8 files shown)


tauri-app/src/App.tsx
```diff
@@ -528,10 +528,18 @@ function AppContent({ username }: AppContentProps) {
             });
 
             try {
-              if (deleteLen > 0) {
+              if (deleteLen > 0 && insertText.length > 0) {
+                await enqueueOp(() =>
```
ChocoChichia: this is my #99 pr problem, rebase and apply to replace too pls

Capture once at the top of the Monaco handler and pass the
snapshotted value:
const baseSeq = lastAppliedSeqRef.current;
// ...
await enqueueOp(() =>
invoke("replace", { position: offset, deleteLength: deleteLen,
content: insertText, baseSeq }),
);

tauri-app/src/App.tsx
ChocoChichia, line 1: on line 541 on await sendDelete(offset, 
deleteLen);:

After rebase on main, sendDelete requires a third baseSeq 
argument (see PR #99's opQueue.ts). This will stop typechecking.
Forward the same captured baseSeq from the handler.

ChocoChichia (changes requested):

ChocoChichia (approved):

PR #104, opened 2026-05-25
Change snapshot exchange to request-response

CarlTeapot:
* removed per-X operation snapshot publishing to gateway
* removed snapshot cache from gateway
* refactored snapshtting to be a synchronous, request-response process when guest joins the room

### DIFF
--- a/gateway/internal/hub/hub_test.go
+++ b/gateway/internal/hub/hub_test.go
@@ -51,6 +51,35 @@ func dial(t *testing.T, url string) *websocket.Conn {
 	return c
 }
 
+func answerSnapshotRequest(t *testing.T, host *websocket.Conn) {
+	t.Helper()
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+	_, request, err := host.Read(ctx)
+	if err != nil {
+		t.Fatalf("host snapshot request read: %v", err)
+	}
+	if len(request) != 2 || request[0] != wire.PrefixControl || request[1] != wire.ControlSnapshotRequest {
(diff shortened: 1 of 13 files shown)


tauri-app/src-tauri/src/state/document/wire_dispatch.rs
```diff
@@ -10,27 +10,49 @@ use crate::state::document::types::DocOp;
...
+            flush_pending_ops(doc_tx, pending_ops).await?;
         }
         return Ok(());
     }
 
     if let Some(op) = decode_op_frame(bytes) {
+        if should_wait_for_snapshot(app, *snapshot_applied) {
+            pending_ops.push(op);
+            return Ok(());
+        }
```
ggela22, line 55: host_commands.rs calls connect() before it calls complete_host(). in the window between those two calls, is_host() returns false. If any op arrives at the host's websocket during that window, should_wait_for_snapshot returns true and the op is pushed into pending_ops. since the host never receives an inbound snapshot frame, snapshot_applied stays false forever, and those ops sit in pending_ops for the lifetime of the connection.

honestly it's probably fine to skip this though since it's basically impossible to cause the bug in reality, no one even has the url at the timestamp the bug is possible at, so we're not gonna get any ops

gateway/internal/hub/hub_test.go
ggela22, line 138: add answerSnapshotRequest / readSnapshot calls, same as you did in the fanout tests. right now it waits 5 seconds for a snapshot reply before it timeouts

ggela22, line 224: add answerSnapshotRequest / readSnapshot calls here too

PR #105, opened 2026-05-27
garbage collection

CarlTeapot:


gateway/internal/wire/wire.go
```diff
@@ -1,21 +1,31 @@
...
 )
 
 const (
 	ControlSessionEnded    byte = 0x01
 	ControlSnapshotRequest byte = 0x02
 )
 
+const (
+	PresenceJoined byte = 0x00
+	PresenceLeft   byte = 0x01
```
CarlTeapot: Confusing naming. Maybe rename it to client join prefix and client leave prefix

gateway/internal/room/room.go
```diff
@@ -155,6 +208,8 @@ func (r *Room) Run() {
 				if !r.deliverSnapshotResponse(msg.Sender, msg.Data) {
 					slog.Debug("snapshot frame dropped: no pending snapshot request", "room_id", r.ID, "client_id", msg.Sender.ID, "bytes", len(msg.Data))
 				}
+			} else if wire.IsGcCommitFrame(msg.Data) {
+				r.broadcast(msg)
```
CarlTeapot: this looks same as the else {} block at the first glance. yeah the function is doing additional validation but maybe we could come up withg something more readable. You can ignore this for now and we can fix this in a separate gateway refactor PR

gateway/internal/room/room.go
```diff
@@ -88,12 +88,65 @@ func (r *Room) Leave(c *client.Client, onEmpty func()) {
...
+		frame, err := wire.EncodePresenceFrame(c.ID, wire.PresenceLeft)
+		if err != nil {
+			slog.Warn("leave: failed to encode presence frame", "room_id", r.ID, "client_id", c.ID, "error", err)
+		} else {
+			for _, peer := range remaining {
+				if !peer.Send(frame) {
+					slog.Warn("leave: slow peer; dropping presence frame", "room_id", r.ID, "peer_id", peer.ID)
+				}
+			}
+			slog.Debug("leave: presence Left broadcast", "room_id", r.ID, "client_id", c.ID, "notified", len(remaining))
```
CarlTeapot: what the fuck

gateway/internal/room/room.go
```diff
@@ -88,12 +88,65 @@ func (r *Room) Leave(c *client.Client, onEmpty func()) {
 		close(r.done)
 		slog.Info("room marked closed after last client left", "room_id", r.ID)
 	}
+	remaining := make([]*client.Client, 0, len(r.clients))
+	for peer := range r.clients {
+		remaining = append(remaining, peer)
+	}
```
CarlTeapot: We could use a read/write lock over the clients map to avoid creating new slices. But you can ignore this too and consider it in the gateway refactor

gateway/internal/room/room.go
```diff
@@ -88,12 +88,65 @@ func (r *Room) Leave(c *client.Client, onEmpty func()) {
...
+			slog.Warn("leave: failed to encode presence frame", "room_id", r.ID, "client_id", c.ID, "error", err)
+		} else {
+			for _, peer := range remaining {
+				if !peer.Send(frame) {
+					slog.Warn("leave: slow peer; dropping presence frame", "room_id", r.ID, "peer_id", peer.ID)
+				}
+			}
+			slog.Debug("leave: presence Left broadcast", "room_id", r.ID, "client_id", c.ID, "notified", len(remaining))
+		}
+	}
```
CarlTeapot: We could move this to a separate function like emitClientLeave or something

gateway/internal/room/room.go
```diff
@@ -88,12 +88,65 @@ func (r *Room) Leave(c *client.Client, onEmpty func()) {
...
+	for _, peer := range peers {
+		peerFrame, err := wire.EncodePresenceFrame(peer.ID, wire.PresenceJoined)
+		if err != nil {
+			slog.Warn("emit-joined: failed to encode existing-peer presence frame", "room_id", r.ID, "peer_id", peer.ID, "error", err)
+			continue
+		}
+		if !newClient.Send(peerFrame) {
+			slog.Warn("emit-joined: slow new client; dropping peer presence frame", "room_id", r.ID, "client_id", newClient.ID)
+		}
+	}
```
CarlTeapot: does new client need these messages? doesn't he already have all existing client info embedded in the snapshot that client receives?

gateway/internal/room/room.go
```diff
@@ -88,12 +88,65 @@ func (r *Room) Leave(c *client.Client, onEmpty func()) {
...
+		} else {
+			for _, peer := range remaining {
+				if !peer.Send(frame) {
+					slog.Warn("leave: slow peer; dropping presence frame", "room_id", r.ID, "peer_id", peer.ID)
+				}
+			}
+			slog.Debug("leave: presence Left broadcast", "room_id", r.ID, "client_id", c.ID, "notified", len(remaining))
+		}
+	}
+}
```
CarlTeapot: Lets add an info log that confirms that leave was successful. I think we will need it

tauri-app/src-tauri/src/session/host_commands.rs
```diff
@@ -27,6 +28,10 @@ pub async fn host_session(app: AppHandle) -> Result<(), String> {
 
     let disconnect_rx = connect(&app, setup.port, setup.room_id.clone()).await?;
 
+    let (gc_tx, gc_rx) = mpsc::channel(64);
+    app.state::<AppState>().set_gc_tx(gc_tx);
+    tauri::async_runtime::spawn(gc_coordinator_loop(gc_rx, app.clone()));
```
CarlTeapot: Maybe this should start after .complete_host() returns successfully, if it fails we have a fuckass job running for no reason

tauri-app/src-tauri/src/state/ws_state.rs
```diff
@@ -81,8 +82,11 @@ impl WsState {
         let (write_tx, write_rx) = mpsc::channel::<Message>(64);
         let (op_tx, op_rx) = mpsc::unbounded_channel::<Vec<u8>>();
         let (disconnect_tx, disconnect_rx) = oneshot::channel::<DisconnectReason>();
+        let (sv_signal_tx, sv_signal_rx) = mpsc::channel::<()>(1);
         debug!("ws channel created: write_buffer_capacity=64");
-        let processor = tokio::task::spawn(process_loop(op_rx, app.clone()));
+        let processor = tokio::task::spawn(process_loop(op_rx, app.clone(), sv_signal_tx.clone()));
+        tokio::task::spawn(sv_publisher_loop(sv_signal_rx, app.clone()));
```
CarlTeapot: Right now this code starts the state vector publisher loop for the host too, which is pointless. It doesn't break anything because guests do not receive it due to not having a garbage collection process running, but its still dead weight. Lets come up with a way to refactor this. I don't think a simple if else block would be good too, it should be decided by upstream entities like app state or app role

crdt-core/src/document/pending.rs
```diff
@@ -73,17 +73,25 @@ impl Document {
...
+            && !self.store.contains_key(&or_id)
+            && !self.origin_observable(or_id)
         {
             return BlockReadiness::Pending;
         }
         BlockReadiness::Ready
     }
 
+    /// Returns true if the block identified by `origin` has been observed in our
+    /// state vector, meaning we have integrated it (or it has been GC'd after integration)
```
CarlTeapot: no need for a comment for a pretty well understandable function. it will need maintenance later. I think we should only keep comments for functions where it is very difficult to implement it in a way that will be easy to understand

crdt-core/src/document/ops.rs
```diff
@@ -1,10 +1,41 @@
...
+) -> Option<BlockId> {
+    let mut current = start;
+    loop {
+        let id = current?;
+        if !removed.contains(&id) {
+            return Some(id);
+        }
+        current = store.get(&id)?.right();
+    }
+}
```
CarlTeapot: why are these functions not part of document impl? are they used by other modules too? if not, I think it would be better to put them in document impl in traversal.rs, not here

crdt-core/src/document/ops.rs
```diff
@@ -179,34 +210,72 @@ impl Document {
...
+            while clock < range.end() {
+                let id = BlockId::new(*client, Clock::new(clock));
                 let Some(block) = self.store.get(&id) else {
                     break;
                 };
-                let next_clock = block.id.clock.value + block.len;
+                clock = block.id.clock.value + block.len;
+                to_remove.insert(id);
+            }
+        }
```
CarlTeapot: Lets move this to a separate function that only flattens all blocks that need to be deleted. it will be more readable and the comment wont be necessary. You could also make it a function of delete set called flatten() or something

crdt-core/src/document/ops.rs
```diff
@@ -179,34 +210,72 @@ impl Document {
...
+                self.store
+                    .get_mut(&l_id)
+                    .unwrap()
+                    .set_right(surviving_right);
+            } else {
+                self.head = surviving_right;
+            }
+            if let Some(r_id) = surviving_right {
+                self.store.get_mut(&r_id).unwrap().set_left(surviving_left);
             }
```
CarlTeapot: What happens to the blocks whose origin left/right block gets garbage collected? then their origin left/right becomes null, Could this break something? Should we set the origin left/right to the surviving left/right blocks?

crdt-core/src/store/delete_set.rs
```diff
@@ -88,6 +88,63 @@ impl DeleteSet {
         self.ranges.values().all(|v| v.is_empty())
     }
 
+    pub fn subtract(&mut self, other: &DeleteSet) {
```
CarlTeapot: maybe incoming_ranges better describes what the parameter represents. "other" doesn't really convey anything

crdt-core/src/store/delete_set.rs
```diff
@@ -88,6 +88,63 @@ impl DeleteSet {
...
+                    i += 1;
+                }
+                (Some(left), Some(right)) => {
+                    list[i] = left;
+                    list.insert(i + 1, right);
+                    i += 2;
+                }
+            }
+        }
+    }
```
CarlTeapot: I think this can be simplified too if we choose to communicate garbage collection commands by just specifying the max clock id that should be deleted, instead of different delete ranges.

tauri-app/src-tauri/src/state/gc_signal.rs
```diff
@@ -0,0 +1,13 @@
...
+#[derive(Debug)]
+pub enum GcSignal {
+    PeerSv {
+        client_id: ClientId,
+        sv: StateVector,
+    },
+    PeerLeft {
+        client_id: ClientId,
+    },
+}
```
CarlTeapot: should this be in a separate file? maybe moving it into the gc_coordinator or types.rs could be a better option

tauri-app/src-tauri/src/ws_management/sv_publisher.rs
```diff
@@ -0,0 +1,69 @@
...
+        tokio::select! {
+            Some(()) = trigger_rx.recv() => {
+                send_sv(&app, &doc_tx, client_id).await;
+                heartbeat.reset();
+            }
+            _ = heartbeat.tick() => {
+                send_sv(&app, &doc_tx, client_id).await;
+            }
+        }
+    }
```
CarlTeapot: This doesnt handle when the websocket processing stops and trigger_rx closes. What do you think about adding a return statement when trigger_rx.recv() returns null?

tauri-app/src-tauri/src/state/ws_state.rs
```diff
@@ -81,8 +82,11 @@ impl WsState {
...
+        let _ = sv_signal_tx.try_send(());
```
CarlTeapot: state vector publisher and gc coordinator jobs should start after reader and writer loops start and the connection is established

tauri-app/src-tauri/src/ws_management/gc_coordinator.rs
```diff
@@ -0,0 +1,135 @@
...
+            return;
+        }
+    };
+    let delete_set = match request(doc_tx, |r| DocOp::GetDeleteSet { reply: r }).await {
+        Ok(ds) => ds,
+        Err(e) => {
+            warn!("gc_coordinator: failed to read delete_set: {e}");
+            return;
+        }
+    };
```
CarlTeapot: What if another operation goes into the document actor channel between these two requests? then you can get mismatching delete set and state vector. I cant think of a specific example that can happen that will break any functionalities, but it would still be better if this was one request. maybe we could combine get delete set and get state vector in a combined, atomic FetchGCData command Doc Op or something

tauri-app/src-tauri/src/ws_management/gc_coordinator.rs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        warn!("gc_coordinator: doc actor channel closed");
+        return;
+    }
+
+    let bytes = encode_gc_commit(&GcCommit {
+        confirmed: confirmed.clone(),
+    });
+    app.state::<WsState>().send_raw(bytes).await;
+
```
CarlTeapot: What if this request fails? Do we have any kind of retry mechanism? if we dont, then we have deleted blocks from the host that arent deleted in other peers. One mitigation could also be that next round of garbage collection would eventually take care of it, but that option can only work if we get rid of the explicit delete set interval sending

CarlTeapot: Currently, we use debounce intervals to send gc commits. The thing is, we already use time intervals for peers when they send state vectors. We could simplify the host gc commit sending with the following idea: 
* Host can keep received boolean flags for each peer. when the host receives SV, received flag is set to true. when host receives SV-s from all peers, it can send GCCommit message and then set the received flags to false. I think this can simplify our code by getting rid of time interval state keeping

tauri-app/src-tauri/src/state/gc_signal.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[derive(Debug)]
+pub enum GcSignal {
+    PeerSv {
+        client_id: ClientId,
+        sv: StateVector,
+    },
+    PeerLeft {
+        client_id: ClientId,
+    },
```
CarlTeapot: We could improve the naming: PeerStateVector and PeerLeftMessage or something similar

tauri-app/src-tauri/src/ws_management/sv_publisher.rs
```diff
@@ -0,0 +1,69 @@
...
+use tokio::time::{self, Duration};
+
+use crate::state::appstate::AppState;
+use crate::state::document::client::request;
+use crate::state::document::types::DocOp;
+use crate::state::ws_state::WsState;
+
+const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
+
+pub async fn sv_publisher_loop(mut trigger_rx: mpsc::Receiver<()>, app: AppHandle) {
```
CarlTeapot: this could have a better name, like sv_request_rx

tauri-app/src-tauri/src/ws_management/gc_coordinator.rs
```diff
@@ -0,0 +1,135 @@
...
+                next_gc = far_future;
+                maybe_emit_gc(&app, &doc_tx, &peer_svs, &mut last_emitted).await;
+            }
+        }
+    }
+}
+
+async fn maybe_emit_gc(
+    app: &AppHandle,
+    doc_tx: &crate::state::document::client::DocSender,
```
CarlTeapot: just import bro

tauri-app/src-tauri/src/ws_management/gc_coordinator.rs
```diff
@@ -0,0 +1,135 @@
...
+        if min_coverage > range.start {
+            let confirmable_len = (min_coverage - range.start).min(range.len);
+            confirmed.add(
+                BlockId::new(*client, Clock::new(range.start)),
+                confirmable_len,
+            );
+        }
+    }
+    confirmed
+}
```
CarlTeapot: I think we can create a better abstraction layer for the entire garbage collection system (gc coordinator and sv publisher).  Right now, the channel creation, the task spawning and other provisioning is scattered across different modules like host_commands.rs and ws management and etc. What do you think about creating a separate module that handles all kinds of provisioning and channel instantiation and task spawning itself? You researched and worked on this so you can probably think of a better way to do this than me, but Codex suggested something like this and maybe that can help get you started:  
pub struct SyncMaintenance {
gc_tx: mpsc::Sender<GcSignal>,
sv_tx: mpsc::Sender<SvPublishReason>,
}

impl SyncMaintenance {
pub fn start_host(app: AppHandle) -> Self;
pub fn start_connection(app: AppHandle) -> SvPublisherHandle;

      pub fn peer_state_vector(&self, client_id: ClientId, sv: StateVector);
      pub fn peer_left(&self, client_id: ClientId);
      pub fn request_sv_publish(&self, reason: SvPublishReason);
}

But I’d probably split the lifetime handles:

pub struct GcCoordinatorHandle {
tx: mpsc::Sender<GcSignal>,
task: JoinHandle<()>,
}

pub struct SvPublisherHandle {
tx: mpsc::Sender<SvPublishReason>,
task: JoinHandle<()>,
}

Then callers stop dealing with raw mpsc::Sender<()> and raw GcSignal.

What This Improves

Right now, invocation is scattered:

- host_commands.rs creates the GC channel and spawns gc_coordinator_loop.
- ws_state.rs creates the SV trigger channel and spawns sv_publisher_loop.
- wire_dispatch.rs knows how to translate wire frames directly into GC coordinator messages.
- AppState stores a raw mpsc::Sender<GcSignal>.

A better boundary is:

sync_maintenance::start_host_gc(app.clone());
sync_maintenance::start_sv_publisher(app.clone());
sync_maintenance::handle_incoming_sync_frame(app, frame);

or, more explicitly:

app.state::<SyncMaintenanceState>()
.peer_state_vector(frame.client_id, sv);

app.state::<SyncMaintenanceState>()
.peer_left(frame.client_id);

app.state::<SyncMaintenanceState>()
.publish_sv(SvPublishReason::SnapshotApplied);

That gives you a dedicated place for debounce policy, channel capacity, lifecycle, logging, and future protocol changes.

CarlTeapot (commented):
tashi

PR #106, opened 2026-05-27
14 optimize block retrieval by text position

### LINKED ISSUE #14: optimize block retrieval by text position

ChocoChichia:
#14 
- Adds crdt-core/src/index/ - a Vec-backed
  augmented B+ tree (PositionIndex) that maintains a
  mapping from CRDT block IDs to visible document
  positions in O(log n).
  - position_of(id) and find_at_position(pos) replace
   the previous O(n) linked-list walks used by
  visible_position_of and
  get_block_and_offset_by_position.
  - The tree is kept in sync with the document at
  every mutation chokepoint (integrate, split_block,
  mark_block_deleted, collect_garbage). A debug-build
   oracle (assert_index_matches_linked_list) fires
  after every mutation to verify the two structures
  agree.
  - Module is S.O.L.I.D-decomposed: structs/ holds
  data-only types, operation files (mutate, find,
  split, propagate, build, descend, storage_ops) each
   own a single concern, and constants.rs holds the
  branching-factor tunables (4 in debug, 32/16 in
  release).
  - 16 targeted index unit tests + 7
  two-client/snapshot reproducer tests for the
  reported cursor-on-line-2 regression; all 92 tests
  pass.

### DIFF
--- a/crdt-core/src/document.rs
+++ b/crdt-core/src/document.rs
@@ -1,3 +1,4 @@
+use crate::index::PositionIndex;
 use crate::store::{DeleteSet, StateVector, StructStore};
 use crate::structs::Block;
 use crate::types::{BlockId, ClientId};
@@ -22,6 +23,7 @@ pub struct Document {
     pub delete_set: DeleteSet,
     pub seen_delete_set: DeleteSet,
     pub head: Option<BlockId>,
+    pub position_index: PositionIndex,
     pending_blocks: Vec<Block>,
     pending_delete_sets: Vec<DeleteSet>,
 }
@@ -35,6 +37,7 @@ impl Document {
             delete_set: DeleteSet::new(),
             seen_delete_set: DeleteSet::new(),
             head: None,
+            position_index: PositionIndex::new(),
             pending_blocks: Vec::new(),
             pending_delete_sets: Vec::new(),
         }
@@ -51,6 +54,20 @@ impl Document {
         pending_blocks: Vec<Block>,
         pending_delete_sets: Vec<DeleteSet>,
     ) -> Self {
+        let mut position_index = PositionIndex::new();
+        let mut entries: Vec<(BlockId, u64, bool)> = Vec::new();
+        let mut curr = head;
+        while let Some(id) = curr {
+            match store.get(&id) {
+                Some(block) => {
+                    entries.push((id, block.len, block.is_deleted));
+                    curr = block.right();
+                }
+                None => break,
+            }
+        }
+        position_index.rebuild_from_order(entries.into_iter());
+
         Document {
             client_id,
             store,
@@ -58,6 +75,7 @@ impl Document {
             delete_set,
             seen_delete_set,
             head,
+            position_index,
             pending_blocks,
             pending_delete_sets,
         }
--- a/crdt-core/src/document.rs
+++ b/crdt-core/src/document.rs
@@ -3,6 +3,8 @@ use crate::store::{DeleteSet, StateVector, StructStore};
 use crate::structs::Block;
 use crate::types::{BlockId, ClientId};
 
+#[cfg(debug_assertions)]
+mod debug;
 mod integrate;
 mod ops;
 mod pending;
@@ -23,7 +25,7 @@ pub struct Document {
     pub delete_set: DeleteSet,
     pub seen_delete_set: DeleteSet,
     pub head: Option<BlockId>,
-    pub position_index: PositionIndex,
+    pub(crate) position_index: PositionIndex,
     pending_blocks: Vec<Block>,
     pending_delete_sets: Vec<DeleteSet>,
 }
--- a/crdt-core/src/document/traversal.rs
+++ b/crdt-core/src/document/traversal.rs
@@ -126,9 +126,17 @@ impl Document {
     /// Visible-text character position of the block identified by `target`.
     /// Backed by the position index: O(log n) instead of an O(n) linked-list walk.
     pub(super) fn visible_position_of(&self, target: BlockId) -> u64 {
-        self.position_index
+        let t0 = std::time::Instant::now();
+        let result = self
+            .position_index
             .position_of(target)
-            .unwrap_or_else(|| self.position_index.visible_len())
+            .unwrap_or_else(|| self.position_index.visible_len());
+        debug!(
+            "visible_position_of: {}µs  doc_blocks={}",
+            t0.elapsed().as_micros(),
+            self.store.total_blocks(),
+        );
+        result
     }
 
     /// Locate the block and intra-block offset that corresponds to a visible
@@ -138,8 +146,13 @@ impl Document {
         &self,
         position: u64,
     ) -> (Option<BlockId>, u64, Option<BlockId>) {
-        debug!("get_block_and_offset_by_position({:?})", position);
+        let t0 = std::time::Instant::now();
         let r = self.position_index.find_at_position(position);
+        debug!(
+            "get_block_and_offset_by_position: {}µs  doc_blocks={}",
+            t0.elapsed().as_micros(),
+            self.store.total_blocks(),
+        );
         (r.id, r.offset, r.tail_id)
     }
 }
--- a/crdt-core/src/document/traversal.rs
+++ b/crdt-core/src/document/traversal.rs
@@ -1,6 +1,6 @@
 use super::Document;
 use crate::types::BlockId;
-use log::{debug, trace};
+use log::trace;
 
 impl Document {
     pub fn get_text(&self) -> String {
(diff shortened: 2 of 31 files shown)


crdt-core/src/document/traversal.rs
```diff
@@ -77,57 +123,36 @@ impl Document {
...
-                break;
-            };
-            if !block.is_deleted {
-                pos += block.len;
-            }
-            curr = block.right();
-        }
-        pos
+        let t0 = std::time::Instant::now();
+        let result = self
```
CarlTeapot: Should we keep this in production?

crdt-core/src/document/traversal.rs
```diff
@@ -77,57 +123,36 @@ impl Document {
...
-            if position < content_len {
-                debug!("position: {}, content_len: {}", position, content_len);
-                return (Some(block.id), position, None);
-            }
-            position -= content_len;
-            current_block = block.right().and_then(|id| self.store.get(&id));
-        }
-
+        let t0 = std::time::Instant::now();
+        let r = self.position_index.find_at_position(position);
```
CarlTeapot: same here

crdt-core/src/document/traversal.rs
```diff
@@ -35,6 +35,52 @@ impl Document {
...
+        assert_eq!(
+            tree_total, pos,
+            "position_index.visible_len() = {}, linked-list visible total = {}",
+            tree_total, pos,
+        );
+
+        self.position_index
+            .debug_validate()
+            .expect("position index invariants must hold");
+    }
```
CarlTeapot: should we keep this in the final version? if so, we should introduce a separate debug.rs file and move it there

crdt-core/src/index/constants.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg(not(debug_assertions))]
+pub(super) const LEAF_CHILDREN: usize = 64;
+#[cfg(not(debug_assertions))]
+pub(super) const NODE_CHILDREN: usize = 32;
+
+#[cfg(debug_assertions)]
+pub(super) const LEAF_CHILDREN: usize = 4;
+#[cfg(debug_assertions)]
+pub(super) const NODE_CHILDREN: usize = 4;
```
CarlTeapot, line 9: why?

docs/b-tree-optimisation.md
```diff
@@ -0,0 +1,413 @@
...
+> Now we have a B+ tree that jumps straight there in O(log n). At 200 blocks that's **9× faster
+> (127µs → 14µs)**. At 10 000 blocks it becomes **340× faster** and the difference is felt.
+
+---
+
+## 1. The problem we had
+
+The CRDT document is a **linked list of blocks**. Each block holds a chunk of text and a pointer
+to the next block.
+
```
CarlTeapot, line 13: We should add a rule in .claude/rules  that imports this doc and restrict the regex pattern to crdt-core/** in the YAML frontmatter so that rule is only loaded when working with crdt

crdt-core/src/index/structs/node.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub visible_len: u64,
+}
+
+#[derive(Debug)]
+pub(in crate::index) struct Node {
+    pub child_slots: [Option<ChildSlot>; NODE_CHILDREN],
+    pub num_children: u8,
+    pub parent: Option<NodeIdx>,
+    pub is_leaf_parent: bool,
+}
```
CarlTeapot: maybe an enum can be more descriptive:
pub child_type: ChildType

crdt-core/src/index/storage_ops.rs
```diff
@@ -0,0 +1,78 @@
...
+        leaf_idx: LeafIdx,
+        after_slot: Option<u8>,
+        entry: LeafEntry,
+    ) {
+        let leaf = &mut self.leaves[leaf_idx.0 as usize];
+        debug_assert!((leaf.num_entries as usize) < LEAF_CHILDREN);
+        let insert_pos = match after_slot {
+            None => 0,
+            Some(s) => s as usize + 1,
+        };
```
CarlTeapot, line 36: maybe you could add the check for after_slot that it cant be bigger than total length so we avoid unecessary for loop and reindexing entries  in the hashmap ?

crdt-core/src/index/storage_ops.rs
```diff
@@ -0,0 +1,78 @@
...
+            leaf.entries[i + 1] = leaf.entries[i].take();
+        }
+        leaf.entries[insert_pos] = Some(entry);
+        leaf.num_entries += 1;
+
+        for slot in 0..leaf.num_entries as usize {
+            if let Some(e) = &leaf.entries[slot] {
+                self.id_to_leaf.insert(e.id, (leaf_idx, slot as u8));
+            }
+        }
```
CarlTeapot: you could reindex entries starting from the insert_pos  instead of the start

crdt-core/src/index/build.rs
```diff
@@ -0,0 +1,53 @@
...
+            idx: right.0,
+            visible_len: right_visible,
+        });
+        node.num_children = 2;
+        let node_idx = NodeIdx(self.nodes.len() as u32);
+        self.nodes.push(node);
+        self.nodes[left.0 as usize].parent = Some(node_idx);
+        self.nodes[right.0 as usize].parent = Some(node_idx);
+        node_idx
+    }
```
CarlTeapot, line 65: bro this code is nearly identical. something like this looks better:
```

pub fn make_root_node_for_two_leaves(
&mut self,
left: LeafIdx,
left_visible: u64,
right: LeafIdx,
right_visible: u64,
) -> NodeIdx {
let node_idx =
self.push_node_with_two_children(true, left.0, left_visible, right.0, right_visible);

          self.leaves[left.0 as usize].parent = Some(node_idx);
          self.leaves[right.0 as usize].parent = Some(node_idx);

          node_idx
      }

      pub fn make_root_node_for_two_nodes(
          &mut self,
          left: NodeIdx,
          left_visible: u64,
          right: NodeIdx,
          right_visible: u64,
      ) -> NodeIdx {
          let node_idx =
              self.push_node_with_two_children(false, left.0, left_visible, right.0, right_visible);

          self.nodes[left.0 as usize].parent = Some(node_idx);
          self.nodes[right.0 as usize].parent = Some(node_idx);

          node_idx
      }

      fn push_node_with_two_children(
          &mut self,
          is_leaf_parent: bool,
          left_idx: u32,
          left_visible: u64,
          right_idx: u32,
          right_visible: u64,
      ) -> NodeIdx {
          let mut node = Node::new(is_leaf_parent);
          node.child_slots[0] = Some(ChildSlot {
              idx: left_idx,
              visible_len: left_visible,
          });
          node.child_slots[1] = Some(ChildSlot {
              idx: right_idx,
              visible_len: right_visible,
          });
          node.num_children = 2;

          let node_idx = NodeIdx(self.nodes.len() as u32);
          self.nodes.push(node);
          node_idx
      }
```

crdt-core/src/index/descend.rs
```diff
@@ -0,0 +1,58 @@
...
+                {
+                    s.visible_len = (s.visible_len as i64 + delta) as u64;
+                    break;
+                }
+            }
+            parent_opt = node.parent;
+            child_idx_u32 = parent.0;
+        }
+    }
+}
```
CarlTeapot, line 49: You could use an iterator instead of a loop to find the slot. also same as before,  common behavior should be moved in a separate function: 
```

pub fn bubble_visible_len_delta(&mut self, leaf_idx: LeafIdx, delta: i64) {
let parent = self.leaves[leaf_idx.0 as usize].parent;
self.bubble_visible_len_delta_from_child(parent, leaf_idx.0, delta);
}

      pub fn bubble_visible_len_delta_from_node(&mut self, node_idx: NodeIdx, delta: i64) {
          let parent = self.nodes[node_idx.0 as usize].parent;
          self.bubble_visible_len_delta_from_child(parent, node_idx.0, delta);
      }

      fn bubble_visible_len_delta_from_child(
          &mut self,
          mut parent_opt: Option<NodeIdx>,
          mut child_idx: u32,
          delta: i64,
      ) {
          while let Some(parent) = parent_opt {
              let node = &mut self.nodes[parent.0 as usize];
              let slot = node
                  .child_slots
                  .iter_mut()
                  .take(node.num_children as usize)
                  .flatten()
                  .find(|slot| slot.idx == child_idx)
                  .expect("parent slot pointing to child not found");

              slot.visible_len = (slot.visible_len as i64 + delta) as u64;
              parent_opt = node.parent;
              child_idx = parent.0;
          }
      }

```

crdt-core/src/index/propagate.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    fn overwrite_parent_slot_for_left(
+        &mut self,
+        parent: NodeIdx,
+        left_idx_u32: u32,
+        new_left_visible: u64,
+        right_visible: u64,
+    ) -> i64 {
+        let node = &mut self.nodes[parent.0 as usize];
```
CarlTeapot, line 103: what the helly

crdt-core/src/document.rs
```diff
@@ -22,6 +23,7 @@ pub struct Document {
     pub delete_set: DeleteSet,
     pub seen_delete_set: DeleteSet,
     pub head: Option<BlockId>,
+    pub position_index: PositionIndex,
```
CarlTeapot: This shouldn't be public. tauri shouldnt have direct access to the position index. maybe the same goes with the head variable too.

crdt-core/src/index/mutate.rs
```diff
@@ -0,0 +1,176 @@
...
+enum InsertTarget {
+    DoneEmpty,
+    At {
+        leaf_idx: LeafIdx,
+        after_slot: Option<u8>,
+    },
+}
+
+impl PositionIndex {
+    pub fn insert_after(&mut self, prev: Option<BlockId>, id: BlockId, len: u64) {
```
CarlTeapot, line 18: we should check if the block id is already present in the index and reject it

crdt-core/src/index/mutate.rs
```diff
@@ -0,0 +1,176 @@
...
+            self.storage.bubble_visible_len_delta(leaf_idx, len as i64);
+            return;
+        }
+
+        let (right_idx, left_visible, right_visible) =
+            self.storage.split_leaf(leaf_idx, after_slot, entry);
+        self.attach_after_leaf_split(leaf_idx, left_visible, right_idx, right_visible);
+    }
+
+    pub fn split_entry(&mut self, id: BlockId, offset: u64, new_id: BlockId) {
```
CarlTeapot, line 49: same here, this should check new_id and verify that it doesnt already exist

crdt-core/src/index/mutate.rs
```diff
@@ -0,0 +1,176 @@
...
+            InsertTarget::At {
+                leaf_idx,
+                after_slot,
+            } => (leaf_idx, after_slot),
+        };
+
+        let leaf = &self.storage.leaves[leaf_idx.0 as usize];
+        if (leaf.num_entries as usize) < LEAF_CHILDREN {
+            self.storage.insert_into_leaf(leaf_idx, after_slot, entry);
+            self.storage.bubble_visible_len_delta(leaf_idx, len as i64);
```
CarlTeapot, line 40: why not unsigned? it will never be negative

crdt-core/src/index/split.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// appropriate slot. Returns the new right-hand leaf's index and the
+    /// left/right visible-len values. The original leaf becomes the left half.
+    pub fn split_leaf(
+        &mut self,
+        leaf_idx: LeafIdx,
+        after_slot: Option<u8>,
+        entry: LeafEntry,
+    ) -> (LeafIdx, u64, u64) {
+        let mut all = self.build_leaf_overflow_buffer(leaf_idx, after_slot, entry);
+        let mid = all.len() / 2;
```
CarlTeapot, line 18: Codex suggests that ArrayVec is more optimal than heap Vec

CarlTeapot (changes requested):

PR #108, opened 2026-06-15
metrics

CarlTeapot:
metrics data polling from cloudflared tunnel metrics server and gateway. appear as popups on frontend

### DIFF
--- a/gateway/internal/hub/hub.go
+++ b/gateway/internal/hub/hub.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"gateway/internal/client"
+	gatewaymetrics "gateway/internal/metrics"
 	"gateway/internal/room"
 	"gateway/internal/wire"
 )
@@ -27,12 +28,17 @@ const maxRoomJoinRetries = 3
 const maxInboundMessageBytes = 64 * 1024 * 1024
 
 type Hub struct {
-	mu    sync.Mutex
-	rooms map[string]*room.Room
+	mu      sync.Mutex
+	rooms   map[string]*room.Room
+	metrics *gatewaymetrics.Registry
 }
 
-func New() *Hub {
-	return &Hub{rooms: make(map[string]*room.Room)}
+func New(registries ...*gatewaymetrics.Registry) *Hub {
+	registry := gatewaymetrics.New()
+	if len(registries) > 0 && registries[0] != nil {
+		registry = registries[0]
+	}
+	return &Hub{rooms: make(map[string]*room.Room), metrics: registry}
 }
 
 func newRoomID() (string, error) {
@@ -65,8 +71,9 @@ func (h *Hub) getOrCreateRoom(id string) *room.Room {
 	defer h.mu.Unlock()
 	r, ok := h.rooms[id]
 	if !ok {
-		r = room.New(id)
+		r = room.New(id, h.metrics)
 		h.rooms[id] = r
+		h.metrics.RoomOpened()
 		go r.Run()
 		slog.Info("room created", "room_id", id)
 	} else {
@@ -80,6 +87,7 @@ func (h *Hub) discardIfStale(id string, r *room.Room) {
 	defer h.mu.Unlock()
 	if h.rooms[id] == r {
 		delete(h.rooms, id)
+		h.metrics.RoomClosed()
 		slog.Info("room discarded from hub", "room_id", id)
 	} else {
 		slog.Debug("discard skipped: room reference is stale", "room_id", id)
@@ -143,8 +151,9 @@ func readWSParams(w http.ResponseWriter, r *http.Request) (roomID, clientID stri
 	return roomID, clientID, true
 }
 
-func dispatchFrame(rm *room.Room, sender *client.Client, raw []byte) {
+func (h *Hub) dispatchFrame(rm *room.Room, sender *client.Client, raw []byte) {
 	if err := wire.ValidateFrame(raw); err != nil {
+		h.metrics.FrameDropped()
 		slog.Warn(
 			"dropping invalid frame",
 			"room_id", rm.ID,
@@ -175,6 +184,7 @@ func dispatchFrame(rm *room.Room, sender *client.Client, raw []byte) {
 			"bytes", len(raw),
 		)
 	case <-t.C:
+		h.metrics.FrameDropped()
 		slog.Warn(
 			"room ops buffer full; dropping frame after timeout",
 			"room_id", rm.ID,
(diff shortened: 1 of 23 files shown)


ChocoChichia (approved):

tauri-app/src-tauri/src/processes/tunnel_process.rs
```diff
@@ -70,6 +82,20 @@ pub async fn run_cloudflared(
     }
 }
 
+fn find_free_metrics_addr() -> Result<SocketAddrV4, String> {
+    let listener = TcpListener::bind((Ipv4Addr::LOCALHOST, 0))
```
ChocoChichia, line 86: Port race here. We grab a free port, then drop the listener when 
 this function returns ,- and only after that do we hand the port to 
 cloudflared. In the gap, another process can steal it. If that 
 happens, cloudflared can't bind --metrics and (depending on 
 version) the whole tunnel dies, not just metrics.
  
Low odds on loopback, but the failure is bigger than "no metrics." 
Can we either keep the listener alive until cloudflared is spawned,
or at least add a comment flagging the known tradeoff?


to translate to xixo language if this is too much : 
  1. Our code asks the OS: "give me any free port" -> OS says "port 5000
  is free"
  2. Our code lets go of port 5000 (so it's free again)
  3. Our code launches cloudflared and says "use port 5000"
  4. cloudflared tries to grab port 5000

gateway/internal/room/room.go
```diff
@@ -226,8 +243,11 @@ func (r *Room) getPeers(exclude *client.Client) []*client.Client {
 func (r *Room) sendToPeers(targets []*client.Client, data []byte, slowClientLogMsg string) {
 	for _, c := range targets {
 		if !c.Send(data) {
+			r.metrics.SlowClientDisconnected()
```
ChocoChichia: This metric counts failed sends, not actual disconnects.
A slow client can trigger multiple send failures before it's finally removed, so one real disconnect may increase `slow_client_disconnects` several times.
The data is fine, but the metric can be misleading. Consider deduplicating it or renaming it to reflect that it measures send failures rather than disconnects.


xixo language translate : 
  When the gateway sends a message to a peer and the peer is too slow
  to receive it, we:
  - bump the counter slow_client_disconnects (+1)
  - kick the client (ForceClose)
The problem: kicking the client isn't instant. The client stays in
the room's list for a moment longer. So if another message goes out
during that moment, it tries to send to the same already-being-kicked
client → fails again → counter +1 again → tries to kick again.

tauri-app/src-tauri/src/processes/process_metrics_aggregator.rs
```diff
@@ -0,0 +1,216 @@
...
+            continue;
+        };
+        let metric_name = series.split_once('{').map_or(series, |(name, _)| name);
+
+        match metric_name {
+            "cloudflared_tunnel_ha_connections" => {
+                ha_connections = Some(metric_u64(metric_name, value)?);
+            }
+            "cloudflared_tunnel_tunnel_register_success" => {
+                register_successes += metric_u64(metric_name, value)?;
```
ChocoChichia: Missing-metric handling is inconsistent. `ha_connections` and `request_errors` fail the entire scrape when absent, while `register_successes` silently defaults to `0`.
That means a cloudflared build that doesn't expose `register_successes` reports zero successes instead of an error, causing the frontend's "Reconnecting" detection to never trigger.
Can we use the same policy for all three metrics, either fail on missing metrics or treat them all as optional?

xixo language translate : 
When we read cloudflared's stats page, we pull out three numbers. The
cloudflared page is just text lines like:

cloudflared_tunnel_ha_connections 1
cloudflared_tunnel_request_errors 2
cloudflared_tunnel_tunnel_register_success 3

Our parser:
- ha_connections → if this line is missing, fail loudly ("metric is
missing", whole read fails)
- request_errors → if missing, fail loudly
- register_successes → if missing, silently uses 0 and pretends
everything's fine

Why that's a problem: there's a feature that watches
register_successes to detect "the tunnel dropped and is now
reconnecting." If a future cloudflared version stops printing that
line, we silently see 0 forever → the "Reconnecting" indicator never 
lights up and nobody ever notices it's broken, because nothing
errors.

The fix is simple: treat all three the same - either all fail loudly
when missing, or all default to 0. Just be consistent.

tauri-app/src-tauri/src/processes/process_metrics_aggregator.rs
```diff
@@ -0,0 +1,216 @@
...
+            }
+            "cloudflared_tunnel_request_errors" => {
+                request_errors = Some(metric_u64(metric_name, value)?);
```
ChocoChichia, line 151: `request_errors` uses a last-value-wins approach, while `register_successes` sums all matching series.

That's fine today because `request_errors` is unlabeled, but if cloudflared ever adds labels and exposes multiple series, only the last value will be counted. The two counters should use the same aggregation strategy to avoid silently undercounting in the future.


no xixo translate here

tauri-app/src/App.tsx
ChocoChichia, line 1: file longer than my

tauri-app/src/App.tsx
```diff
@@ -390,6 +542,13 @@ function AppContent({ username }: AppContentProps) {
     setLanUrl(null);
     setPublicUrl(null);
     setProcessesRunning({ gateway: "Disabled", tunnel: "Disabled" });
+    setShowGatewayMetrics(false);
+    setShowTunnelMetrics(false);
+    setGatewayMetrics(null);
+    setGatewayMetricsError(null);
+    setRecentGatewayReplayFailures(0);
+    setRecentGatewayDroppedFrames(0);
+    previousGatewayMetricsRef.current = null;
```
ChocoChichia, line 551: Gateway state is cleared here, but tunnel state is cleared elsewhere in a separate effect.

It works today, but the asymmetry is fragile: if that effect's dependencies change, stale tunnel data could persist. Consider handling both gateway and tunnel cleanup in the same place so the reset behavior is consistent and less error-prone.

gateway/internal/hub/hub.go
```diff
@@ -27,12 +28,17 @@ const maxRoomJoinRetries = 3
...
-	mu    sync.Mutex
-	rooms map[string]*room.Room
+	mu      sync.Mutex
+	rooms   map[string]*room.Room
+	metrics *gatewaymetrics.Registry
 }
 
-func New() *Hub {
-	return &Hub{rooms: make(map[string]*room.Room)}
+func New(registries ...*gatewaymetrics.Registry) *Hub {
```
ChocoChichia: Nit: this variadic-optional pattern (also used in `room.go:45`) creates a temporary `Registry` on every call, only to discard it in the common case.

It seems to exist mainly to preserve the zero-arg `New()` API for tests. A required `*Registry` parameter would be simpler, avoid the unnecessary allocation, and make the dependency explicit (tests can just pass `gatewaymetrics.New()`). Not blocking. you decided

ChocoChichia (changes requested):

tauri-app/src/App.tsx
CarlTeapot: your what??

tauri-app/src-tauri/src/processes/tunnel_process.rs
CarlTeapot: not reading all that

tauri-app/src/App.tsx
CarlTeapot: lets put this aside for frontend refactor

ChocoChichia (approved):

